PR #1, opened 2026-07-01
Reference Metal compute backend

digitaloten:
## What

Adds `infr-metal`, a correctness-first implementation of the `Backend` trait — the same seam `infr-cpu` and `infr-vulkan` implement — on Apple Metal. Models now run on the Apple GPU through the backend-agnostic `Graph`, with no coopmat/portability workarounds needed (MoltenVK exposes neither `VK_KHR_cooperative_matrix` nor the portability enumeration the Vulkan path assumes, so the existing Vulkan backend doesn't come up on macOS).

## Design (perf is explicitly out of scope)

- **Host-orchestrated, GPU-computed.** Graph tensors are staged as f32 exactly like the CPU interpreter (`bytes_to_f32`, `in_place_inputs`, output write-back all mirror `infr-cpu`), and each op's arithmetic runs in a small MSL compute kernel over f32 `MTLBuffer`s. Mirroring the proven CPU dataflow is what keeps it in numeric parity.
- **No quant kernels.** Quantized `Op::Linear` weights are dequantized to f32 on the host (reusing `infr_gguf::dequant::dequant_block`) and cached as device buffers, so the shaders stay format-agnostic — one `linear_f32` kernel covers f32/f16 and every GGUF quant type.
- **Unified memory.** `StorageModeShared` buffers make `upload`/`download` a `memcpy`; one command buffer per op (`commit` + `wait`) guarantees strict graph-order execution with no manual barriers. MSL is compiled at init with fast-math disabled for tighter transcendental parity.

## Ops — full coverage

All 15 graph ops implemented and parity-tested against `infr-cpu`:

- **GPU kernels:** `RmsNorm`, `QkNorm`, `Linear` (f32/f16/quant), `Rope`, `QkNormRope`, `Attention` (GQA + causal & sliding-window, online softmax), `GatedAct`, `Add`, `Scale`, `Softcap`, and the per-expert `MoeFfn` matvecs.
- **Host-side** (data movement / sequential per-token recurrences, like the CPU reference): `Copy`, `WriteKv`, `MoeFfn` router+softmax+top-k, `Conv1dSilu`, `DeltaNet`.

This covers every model family the CPU reference supports: dense (Llama / Qwen3 / Gemma), `qwen3moe` (MoE), and `qwen35` / Qwen3-Next (gated DeltaNet).

## Tests

`crates/infr-metal/tests/parity.rs` runs each op on `infr-cpu` (the trusted reference) and on `infr-metal` and asserts the outputs match; the quantized-Linear test compares against a dequant-then-matmul reference to isolate the quant-weight path. **23 tests pass.**

## Integration

Wired into the CLI behind `INFR_METAL=1`, mirroring the existing `INFR_CPU` / Vulkan routing:
- dense + MoE via a new `generate_dense_metal` over the shared `generate_dense_backend` runner;
- Qwen3-Next via a refactor of `qwen35::generate_cpu` into a backend-generic `generate_seam` + `generate_cpu`/`generate_metal` wrappers.

Verified end-to-end on an Apple M3 Pro:

```
$ INFR_METAL=1 INFR_TEMP=0 infr run unsloth/Qwen3-0.6B-GGUF:Q4_K_M "What is the capital of France? Answer in one word."
[metal backend — dense/MoE forward on Apple GPU via the agnostic compute graph (reference)]
<think>
Okay, the user is asking for the capital of France in one word. Let me think. France's capital is Paris. ...
```

Greedy output matches the `INFR_CPU` reference token-for-token until a late near-tie (expected: the CPU path uses int8-quantized activation dots for k-quant weights, this backend uses full-f32 dequant). MoE and Qwen3-Next are covered by the op parity tests (end-to-end needs the large 30B/80B checkpoints).

## Notes / follow-ups

- Requires only the Metal runtime (present on all macOS); no Vulkan SDK / MoltenVK.
- Future work: keep tensors resident on-device and fuse kernels (drop the per-op host round-trip), MSL quant kernels, and moving the MoE/recurrent ops fully onto the GPU.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,12 +17,14 @@ env:
   # to surface upstream incompatibility early. Mirrors the krypt / tikr baseline.
   FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: "true"
 
-# Linux-only: infr is a Vulkan-first engine. `glslc` (shaderc) compiles the
+# Mostly Linux: infr is a Vulkan-first engine. `glslc` (shaderc) compiles the
 # compute shaders at build time, and the dp4a shaders need a recent glslc
 # (GL_EXT_integer_dot_product) — Ubuntu 24.04 ships shaderc 2023.8 which is too
 # old, so we pin ubuntu-26.04 (shaderc 2025+). The GPU integration tests are
 # #[ignore]'d (they need a real Vulkan device); CI builds + runs the CPU suite.
 # ash loads Vulkan at runtime via dlopen, so no libvulkan is needed to build.
+# The `test-macos` job additionally builds on Apple Silicon and runs the
+# reference Metal backend's parity suite against a real Metal device.
 
 jobs:
   fmt:
@@ -87,3 +89,24 @@ jobs:
       - run: rustup update --no-self-update stable
       - uses: Swatinem/rust-cache@v2
       - run: cargo build --release --workspace --locked
+
+  # macOS / Metal: the Linux jobs above never build `infr-metal` (it is
+  # `#![cfg(target_os = "macos")]`). This job builds the workspace on an
+  # Apple-Silicon runner — which has a real Metal device — and runs the
+  # infr-metal parity suite. Those tests are `#[ignore]`d like the Vulkan GPU
+  # tests, so they need an explicit `--include-ignored`.
+  test-macos:
+    name: cargo test (macOS / Metal)
+    runs-on: macos-15
+    steps:
+      - uses: actions/checkout@v7
+      - name: Install glslc (shaderc)
+        run: brew install shaderc
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: stable
+          cache: false
+      - run: rustup update --no-self-update stable
+      - uses: Swatinem/rust-cache@v2
+      - run: cargo build --workspace --locked
+      - run: cargo test -p infr-metal --locked -- --include-ignored
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -196,6 +196,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b4388bee8683e3d04af747c73422af53102d2bd24d9eadb6cbc100baef4b43f8"
 
 [[package]]
+name = "block"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0d8c1fef690941d3e7788d328517591fecc684c084084702d6ff1641e993699a"
+
+[[package]]
 name = "block-buffer"
 version = "0.10.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -342,6 +348,33 @@ dependencies = [
  "once_cell",
  "unicode-width",
  "windows-sys 0.59.0",
+]
+
+[[package]]
+name = "core-foundation"
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b2a6cd9ae233e7f62ba4e9353e81a88df7fc8a5987b8d445b4d90c879bd156f6"
+dependencies = [
+ "core-foundation-sys",
+ "libc",
+]
+
+[[package]]
+name = "core-foundation-sys"
+version = "0.8.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "773648b94d0e5d620f64f280777445740e61fe701025087ec8b57f45c791888b"
+
+[[package]]
+name = "core-graphics-types"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3d44a101f213f6c4cdc1853d4b78aef6db6bdfa3468798cc1d9912f4735013eb"
+dependencies = [
+ "bitflags",
+ "core-foundation",
+ "libc",
 ]
 
 [[package]]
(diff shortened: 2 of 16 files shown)


mxaddict:
## Review — Reference Metal compute backend

Nice work — the crate mirrors the CPU interpreter's dataflow cleanly, the `Backend` seam impl is correct, the build/CI gating is *almost* airtight, and the op parity tests are meaningful. One blocking CI issue, plus a few should-fix items. Details with `file:line` below.

### 🔴 Blocking — breaks Linux CI

**`crates/infr-metal/tests/backend.rs` is missing the `#![cfg(target_os = "macos")]` guard** that `tests/parity.rs:6` has. On Linux `infr-metal` compiles to an empty lib (the crate is `#![cfg(target_os = "macos")]`), so `use infr_metal::MetalBackend;` (backend.rs:4) is an `E0432 unresolved import`. The Linux CI jobs `cargo clippy --workspace --all-targets --locked -- -D warnings` and `cargo nextest run --workspace --locked` both compile this test target and will fail. Reproduced locally:

```
$ cargo clippy -p infr-metal --all-targets
error[E0432]: unresolved import `infr_metal::MetalBackend`
 --> crates/infr-metal/tests/backend.rs:4:5
```

Fix — add the guard as the first attribute (matching `parity.rs`):

```rust
//! Backend-plumbing tests: device init + buffer memory roundtrip.
#![cfg(target_os = "macos")]

use infr_core::backend::{Backend, BufferUsage};
```

### 🟠 Should fix

1. **`assert!(hd <= 256)` (exec.rs:479) is a hard panic, not a graceful error.** The Attention kernel's `float acc[MAX_HD]` caps head_dim at 256 (shaders.rs:217,234). Gemma 3 is exactly 256 (OK), but Gemma 4's full-attention layers use head_dim 512 — a "supported" family per the README would `panic!` rather than erroring cleanly. Return an `Err` (or clamp/validate upstream) instead of asserting.

2. **Add quant-Linear parity tests for the common formats.** `linear_q8_0_matches_dequant_reference` (parity.rs:297) only tests **Q8_0**, and only against a dequant-then-matmul reference — not against `CpuBackend`. **Q4_K / Q5_K / Q6_K** — the formats every real checkpoint ships — have zero test coverage on the Metal dequant path (`weight_buf` → `dequant_block`, exec.rs:180-196). Please add at least a Q4_K and Q6_K case.

3. **Tighten two overstated parity claims** (the code is fine; the wording isn't):
   - Quantized `Op::Linear` does **not** mirror CPU op-for-op: Metal does a full-f32 dequant dot, while CPU quantizes the *activation* to Q8 and does integer dots (lib.rs:2411-2425). Metal is actually the *more* accurate of the two — you already disclose this in the PR body's "late near-tie" note, so just soften the "mirrors op-for-op" phrasing in the `lib.rs`/`parity.rs` doc comments.
   - MoE expert selection is **not** guaranteed bit-identical to the CPU oracle: the router logits use a different f32 summation order (naive sum exec.rs:534-538 vs CPU's 8-accumulator `dot`), so top-k can diverge on a near-tie. Structurally identical and fine in practice, but reword the commit's "bit-identical" claim.

### 🟡 Optional / follow-up

- **`weight_cache` keyed by buffer address, never invalidated** (exec.rs:187, lib.rs:57). Safe today — each `generate_*` builds a fresh `MetalBackend`, so the cache lives one generation and distinct weights = distinct addresses. But it caches *all* weights (not just norms like the CPU path), so if a backend instance is ever reused across models with buffer free/realloc, a recycled address returns stale f32. Worth a `// single-generation lifetime` note or a more robust key (e.g. `(addr, dtype, len)`).
- **`probs[b].partial_cmp(&probs[a]).unwrap()` (exec.rs:546)** panics on a NaN logit. Finite dequantized weights won't produce one, but `partial_cmp(...).unwrap_or(Equal)` is cheap insurance.
- **Untested behaviors** worth a line each eventually: MoE with quantized experts, `QkNormRope` with `freq_factors`, f32 (vs f16) KV cache, and any multi-op graph chaining (every parity test is a single op).

### ✅ Verified correct

- Attention GQA grouping + causal/sliding-window mask (`lo` bound) + online-softmax numerics all match `infr-cpu` bit-for-bit in structure (shaders.rs:228-249 vs lib.rs:2748-2791); Rope/QkNormRope NEOX pairing + theta + freq-factors; RmsNorm/QkNorm eps & mean-of-squares; GatedAct (activation codes, gate/up order, `up_off` stride); Softcap.
- Conv1dSilu + DeltaNet host recurrences are line-for-line faithful to the CPU reference (`h % nk` tiling, softplus, L2+eps, qscale, history shift).
- f32 buffer sizing is consistent across alloc/upload/dispatch (element counts, `dispatch_threads` exact grid, every kernel guards `gid >= n`); params structs are padding-free and packed in field order.
- The `generate_seam` refactor of `qwen35.rs` is behavior-preserving for the existing CPU path; all Metal entry points are `#[cfg(target_os = "macos")]`-gated with a clear off-macOS runtime error; the macOS CI job (real Metal device + `--include-ignored`) is the right shape.

Once the CI guard is added (and ideally the Gemma-4 head_dim assert + a Q4_K/Q6_K test), this is good to merge.

digitaloten:
Thanks for the thorough review! Addressed in 6d09860:

**🔴 Blocking (already in `372aea2` before this)**
- `tests/backend.rs` now carries `#![cfg(target_os = "macos")]`, matching `parity.rs` — Linux CI builds the empty lib cleanly.

**🟠 Should-fix**
1. **head_dim > 256 no longer panics** — `Op::Attention` returns `Err(Error::Unsupported(..))` instead of `assert!`, so a Gemma-family 512 head_dim errors cleanly (exec.rs).
2. **Q4_K + Q6_K quant-Linear parity tests added** — `linear_q4k_matches_dequant_reference` / `linear_q6k_matches_dequant_reference`, dequant-reference comparison like the Q8_0 case. Added `synth_q4k`/`synth_q6k` (well-formed blocks, finite f16 scales) and a shared `check_quant_linear_parity` helper. Full suite: **25 passed** on M3 Pro.
3. **Softened the overstated wording** — lib.rs doc now states quant `Linear` is a full-f32 dequant dot here vs CPU's int8-activation dots (Metal is the more accurate path); MoE router comment notes the naive-sum vs 8-accumulator-dot ordering difference that can flip top-k on a near-tie.

**🟡 Optional (done while in there)**
- MoE top-k sort hardened against NaN: `partial_cmp(..).unwrap_or(Ordering::Equal)`.
- `weight_cache` documented as single-generation lifetime (fresh `MetalBackend` per generation; don't reuse across models).

`cargo clippy -p infr-metal --all-targets` clean, `cargo fmt` applied.

digitaloten:
### Perf note (naive reference — for context, not a target)

Measured on an Apple M3 Pro, `unsloth/Qwen3-0.6B-GGUF:Q4_K_M`, greedy, same prompt:

| Backend | Prefill | Decode |
|---|---|---|
| Metal (this PR, naive) | 8 tok/s | 3.1 tok/s |
| CPU reference (`INFR_CPU`) | 171 tok/s | 19.4 tok/s |
| ratio | ~21× slower | ~6× slower |

The Metal path is currently **slower than the CPU reference** — it's GPU-starved, not GPU-bound, exactly as the "perf is out of scope" design intends:

- one command buffer per op with `commit()` + `wait_until_completed()` → a full CPU↔GPU sync barrier every op, GPU idle in between;
- every op's result is copied host↔device (`read_f32` / `f32_buf`) with no on-device residency.

Prefill takes the worst hit (~21×) because the per-op serialize + host round-trip cancels the many-tokens-per-forward batching win. Greedy output is identical on both paths, so correctness (the goal here) holds.

The listed follow-ups — resident on-device tensors, fused kernels, dropping the per-op round-trip — are exactly what closes this gap.

PR #2, opened 2026-07-01
fix(vulkan): size flash-attention tile to device shared-memory limit

junjitree:
## Problem

Running any hd=128 model on an **NVIDIA GPU** crashes on the first compute submit:

```
Error: backend: queue_wait_idle: The logical device has been lost.
```

The flash-attention prefill shaders (`attn_flash`, `attn_flash_warp`,
`attn_flash_partial`) statically allocate **~58 KB of shared memory** (`Ss` 16 KB
+ `Ps` 8 KB + `Os[64×128]` f32 **32 KB** + softmax state, at `BM=64`). That fits
AMD RADV's 64 KB `maxComputeSharedMemorySize`, but **exceeds it on every other
mainstream Vulkan device**:

| Device | `maxComputeSharedMemorySize` | 58 KB flash tile |
| --- | --- | --- |
| AMD RADV | 65536 (64 KB) | ✅ |
| Intel ANV | 65536 (64 KB) | ✅ |
| **NVIDIA** | **49152 (48 KB)** | ❌ device-lost |
| **MoltenVK / mobile** | ~**32768 (32 KB)** | ❌ device-lost |
| Vulkan spec minimum | 16384 (16 KB) | ❌ |

The limit was never queried, so the over-committed pipeline was created and then
faulted the GPU at dispatch. Flash silently only ever worked on AMD; `INFR_NO_FLASH`
was the only escape hatch on NVIDIA.

## Fix

Detect `maxComputeSharedMemorySize` and pick the largest flash tile that fits:
**BM=64** (58112 B) on 64 KB devices, **BM=32** (29056 B) on NVIDIA / MoltenVK.
`BM` only scales the query-row tile, workgroup size, and shared arrays — `BN` and
the coopmat fragment counts are fixed, so the online-softmax accumulation order is
**identical across tiles**. If even BM=32 won't fit (<29 KB), the transformer falls
back to the non-flash path instead of tripping device-lost.

- **shaders:** `attn_flash{,_warp,_partial}` take a `BM_TILE` build define → `*_bm32` SPIR-V
- **core:** `Capabilities` gains `max_shared_memory_bytes`
- **vulkan:** the recorder selects the tile + workgroup count from the device limit
- **llama:** `use_flash` is gated on the smallest tile fitting the device
- **examples/limits_probe:** prints the shared-memory limit + fp16 coopmat config

## Validation (RTX 2080 Ti, 48 KB shared)

- `attention_prefill_flash_matches_cpu` passes token-for-token vs the CPU reference
  — all shapes incl. hd=64 and the split-K path (previously device-lost on the first case).
- `infr run` / `infr bench` now work **with flash enabled by default** (no
  `INFR_NO_FLASH`): coherent output, prefill pp512 ≈ 3940 t/s, pp2048 ≈ 2387 t/s.
- `cargo fmt` / `clippy` clean. No change on AMD (BM=64 path unchanged).

## Notes

- The opt-in `attn_flash_reg` path (`INFR_FLASH_REG`, `BR=128`) is not on any
  default route and is left untiled — a follow-up if that experiment is kept.

### DIFF
--- a/crates/infr-core/src/backend.rs
+++ b/crates/infr-core/src/backend.rs
@@ -24,6 +24,10 @@ pub struct Capabilities {
     pub f16: bool,
     pub cooperative_matrix: bool,
     pub max_buffer_bytes: u64,
+    /// `maxComputeSharedMemorySize` — the per-workgroup shared-memory budget. Vulkan only guarantees
+    /// 16 KB; RADV gives 64 KB, NVIDIA 48 KB, MoltenVK/mobile often 32 KB. The flash-attention tile
+    /// height is picked to fit this (and flash is skipped entirely if even the smallest tile won't).
+    pub max_shared_memory_bytes: u32,
     pub unified_memory: bool,
     /// The backend records a single-token decode graph ONCE and replays it per token, reading the
     /// position from a device-side params buffer (Vulkan seam) instead of the graph's baked `pos`.
--- a/crates/infr-cpu/src/lib.rs
+++ b/crates/infr-cpu/src/lib.rs
@@ -2248,6 +2248,7 @@ impl Backend for CpuBackend {
             f16: true,
             cooperative_matrix: false,
             max_buffer_bytes: u64::MAX,
+            max_shared_memory_bytes: u32::MAX, // scalar interpreter: no shared-memory tiling
             unified_memory: true,
             // The interpreter reads the baked `pos`/`kv_len` from the graph ops, so the decode graph
             // must be rebuilt per token — no record-once replay.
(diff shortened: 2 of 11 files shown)


junjitree:
### Perf — RTX 2080 Ti (48 KB shared, BM=32 tile), Qwen3-0.6B-Q4_K_M

Before this PR every hd=128 model device-lost on the first submit; the only way to run on NVIDIA was `INFR_NO_FLASH=1`. With the fix, flash runs by default. Prefill t/s (`infr bench -p N -n 0`):

| prefill | flash (this PR) | `INFR_NO_FLASH=1` |
| --- | --- | --- |
| pp512  | ~3915 | ~3880 |
| pp2048 | ~3550 | ~1910 |
| pp8000 | ~1045 | ~1315 |

Flash is ~1.8× at pp2048; decode tg64 ≈ 80 t/s. One honest note: with the smaller BM=32 tile the non-FA path still edges flash at pp8000 (~1315 vs ~1045) — a data point for the existing "auto-select from device bandwidth/FLOP caps" TODO, since the tradeoff shifts once the tile shrinks. Numbers are approximate (single GPU shared with the desktop compositor).

PR #3, opened 2026-07-01
fix(vulkan): size register-O flash tile to device shared-memory limit

junjitree:
Follow-up to #2. A shared-memory audit of every compute shader found one more over the budget — `attn_flash_reg`, the opt-in `INFR_FLASH_REG` register-O prefill:

| shader | shared | NVIDIA 48 KB | MoltenVK 32 KB |
| --- | --- | --- | --- |
| **`attn_flash_reg`** (BR=128) | **58,880 B** | ❌ | ❌ |
| `attn_qk_warp` (next largest) | 25,600 B | ✅ | ✅ |
| everything else | ≤ 25,600 B | ✅ | ✅ |

`sfsh`(32 KB) + `Psh`(16 KB) + `pvsh`(8 KB) + state = 58,880 B. It's reachable only via the `INFR_FLASH_REG` env flag (off every default path), so it doesn't hit normal use — but setting that flag on NVIDIA/Apple device-losts exactly like the default flash path did before #2.

## Fix

Same approach as #2: parametrize the tile height `BR` with a `BR_TILE` build define → `attn_flash_reg_br64` SPIR-V (**29,440 B**, fits both NVIDIA and MoltenVK). `BR` only scales `WARPS_M`, workgroup size, and shared arrays — the per-warp 32×32 coopmat work (`sacc[2][2]`) and `BC` are fixed, so results are identical across tiles. The recorder picks BR=128 (≥58,880) else BR=64 from `maxComputeSharedMemorySize`; the transformer skips reg below 29,440 B. `mpad` stays 128-aligned, so scratch buffers are valid for either tile.

## Validation (RTX 2080 Ti, 48 KB)

- `attention_prefill_flash_reg_matches_cpu` passes vs the CPU reference (now BR=64; previously device-lost).
- `INFR_FLASH_REG=1 infr bench` runs end-to-end (pp2048 ≈ 3428 t/s) — was device-lost.
- `cargo fmt` / `clippy` clean. BR=128 path (AMD ≥64 KB) unchanged.

This closes the last shader over the shared-memory budget found in the audit.

### DIFF
--- a/crates/infr-llama/src/transformer.rs
+++ b/crates/infr-llama/src/transformer.rs
@@ -931,14 +931,12 @@ impl Llama {
         // (hd 256/512, nkv 8/1, wv=None → V=K, E2B per-layer FFN/KV-sharing) via the same per-layer
         // args the GEMV path used — see the QKV/FFN blocks below. INFR_NOGEMM restores GEMV.
         let use_gemm = c.qk_norm && n >= 64 && std::env::var("INFR_NOGEMM").is_err();
-        // Register-O flash (FlashAttention-2 layout, Br=128) is opt-in (INFR_FLASH_REG) while it's
-        // A/B'd vs the BM=64 flash; it needs mpad padded to 128 (q/attn/scratch).
-        // The register-O tile statically allocates 58880 B of shared memory (BR=128); skip it when
-        // the device can't fit that (NVIDIA 48 KB / MoltenVK) so it falls back to the size-aware
-        // flash path instead of over-committing shared memory → device-lost. RADV (64 KB) fits.
+        // Register-O flash (FlashAttention-2 layout, Br=128/64) is opt-in (INFR_FLASH_REG) while it's
+        // A/B'd vs the BM=64 flash; it needs mpad padded to 128 (q/attn/scratch). Its smallest tile
+        // (BR=64) needs 29440 B shared — skip it on devices that can't fit that (else device-lost).
         let use_flash_reg = use_gemm
             && hd == 128
-            && self.be.max_shared_memory_bytes() >= 58880
+            && self.be.max_shared_memory_bytes() >= 29440
             && std::env::var("INFR_FLASH_REG").is_ok();
         // gemma prefill attention: the flash warptile is hd=128-only, but the non-FA coopmat path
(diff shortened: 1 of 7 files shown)


mxaddict:
Reviewed — math checks out (BR*460 shared, thresholds consistent, subgroup geom scales for both tiles). Two polish nits, not blockers:

- **Cap source mismatch risk**: transformer gates on `self.be.max_shared_memory_bytes()` (infr-llama trait); recorder picks BR from `self.be.shared.caps.max_shared_memory_bytes` (infr-vulkan). Confirm both resolve to the same field — if they ever diverge, transformer green-lights reg while recorder sizes wrong.
- **Magic `29440`/`460` in 3 sites**: threshold duplicated as `29440` (transformer) and `128*460`/`64*460` (recorder), all tied to the shader `BR*460`. Add a shared `const` (e.g. `REG_SHARED_PER_BR = 460`) so they can`\t drift if a shared array is added.

LGTM otherwise.

junjitree:
Both nits addressed (pushed + rebased on latest `main`):

1. **Cap source unified** — the recorder now reads the limit via `VulkanBackend::max_shared_memory_bytes()`, the same accessor the transformer uses, in both `attention_prefill_flash` and `attention_prefill_flash_reg`. One field, no divergence path.
2. **Named constants** — added `FLASH_SHARED_PER_ROW = 908` and `FLASH_REG_SHARED_PER_ROW = 460` in `infr-vulkan`, used by both the recorder (tile pick) and the transformer (`use_flash`/`use_flash_reg` gates). Thresholds are now `32 * FLASH_SHARED_PER_ROW`, `64 * FLASH_SHARED_PER_ROW`, `64 * FLASH_REG_SHARED_PER_ROW`, `128 * FLASH_REG_SHARED_PER_ROW` — no bare `29056`/`29440`/`58112`/`58880` left in code.

Rebased onto `6a13a10` (gemma4 GEMM projections). Kept your `INFR_FLASH_BM=32` force flag and the simplified `INFR_NO_FLASH_WARP` gate from that commit; merged the reg guard down to the BR=64 threshold (29440) since this PR makes that tile fit. `fmt`/`clippy` clean; both `attention_prefill_flash{,_reg}_matches_cpu` pass on the 2080 Ti.

PR #4, opened 2026-07-01
perf(metal): profiling-driven Metal backend optimization — decode ~40x, prefill 2.2x

digitaloten:
Follow-up to #1 (reference Metal backend). That backend was correctness-first: every op did a CPU↔GPU round-trip and its own command-buffer commit+wait. This PR makes it fast while keeping exact parity with the CPU reference — every change profiling-driven, A/B-measured back-to-back, and parity-tested (32 GPU tests). Negative results are kept in the history and code comments so they don't get re-attempted blind.

## Results (M3 Pro, Qwen3-0.6B Q4_K_M, greedy)

| metric | before | after |
|---|---|---|
| decode (360-tok run) | ~3.1 tok/s | **~126 tok/s (~40×)** |
| prefill (889-tok prompt) | — / 399 tok/s mid-PR | **~875 tok/s** |
| GPU per decode forward | ~90 ms | **~8.4 ms** |

## What changed, in profiling order

1. **Device residency + batched command buffers** — tensors stay on-device across ops; one commit+wait per forward instead of per op (was 90% of wall).
2. **Simdgroup GEMV + norms** — one simdgroup per output/row with `simd_sum` reduction; coalesced weight reads.
3. **Native-quant weights** — quantized `Op::Linear` stays quantized on device, decoded inline by the kernel; later refactored to the **factored form**: bit-packed 4/6/8-bit codes + i16 `(sc, m)` per 16 elems + f16 `(d, dmin)` per quant block (~6.1 bpw Q4_K / ~8.1 Q6_K vs 32 for f32). `infr-gguf` grew `dequant_factored` as the single affine decoder; `dequant_unified` is its expansion — bit-identical, pinned by the existing llama.cpp-derived tests.
4. **Split-KV attention** — decode's rows=1 launched only `n_head` simdgroups; now NSG simdgroups per (query, head) merged in threadgroup memory, 32-way at long context. Routed for prefill too (measured win).
5. **simdgroup_matrix prefill GEMM** — 8×16 output tile per simdgroup; lanes decode quant blocks into a threadgroup tile, four 8×8×8 MMAs consume it. Prefill Linear 3.5×.
6. **Per-pipeline thread-cap gating** — wide-threadgroup kernels degrade to narrower ones when `maxTotalThreadsPerThreadgroup` is below their need (register pressure, paravirtual CI devices) instead of silently clamping into wrong results.

## Negative results (documented, reverted)

- Row-tiled GEMV (weight reuse across rows): zero change — SLC already absorbed re-reads; the scalar pipeline's 1:1 load:FMA ratio is the real limit (hence MMA).
- f32 simdgroup_matrix flash attention (3 tuning iterations) and a chunked-SIMD variant: never beat the 32-way split on prefill.
- Fused Add+RmsNorm, Linear threadgroup-width sweep, uint4-wide loads: no measured gain.
- Dispatch/barrier overhead measured at ~1.2 µs/kernel (probe test committed) — op fusion is not a lever on this backend.

## Evidence tooling (committed)

- `INFR_METAL_PROFILE=1` (per-op encode + GPU wall) / `=2` (per-op GPU attribution).
- Ignored probe tests: dispatch overhead, device read-bandwidth ceiling (~133 GB/s), quantized GEMV achieved bandwidth (92–106 GB/s).

Bit-for-bit invariant: quantized Linear reconstructs the exact `dequant` reference values (same f32 multiplies); the backend dots in f32 (slightly more accurate than the CPU's Q8-activation integer path, same as before this PR).


digitaloten:
### Update: quant matvec throughput (prefill +76%)

Pushed `556ba95` — two GEMV changes found by continuing to profile:

- **`uchar4`-vectorized code loads** in `linear_qui`: one 4-byte load per lane instead of four 1-byte loads (consecutive lanes cover a full 128-byte cache line/step), 4 weights decoded from a single `(scale,min)` load. Exact same reconstruction — parity unchanged.
- **One-simdgroup threadgroups** (32-wide) for the GEMV, so a matvec launches `out_f` threadgroups instead of a few wide ones → far more for the GPU cores to interleave and hide memory latency.

Measured back-to-back on an M3 Pro (same thermal state — prefill tok/s drifts a lot otherwise), Qwen3-0.6B-Q4_K_M, 614-token prefill:

| | prefill | decode |
|---|---|---|
| before (this PR's first commit) | 107 tok/s | 6.4 |
| after | **188 tok/s (+76%)** | 6.5 |

Decode is flat — it's **dispatch-count bound** (~480 tiny serial GPU dispatches/forward), not matvec bound, so kernel-level matvec tuning doesn't move it. Op fusion (QKV / gate-up, fewer dispatches) is the decode lever — a larger, separate change.

Negative result worth recording: an m-row **tiling** of the GEMV (reuse a decoded weight across TM token-rows) was tried and **reverted** — it cut simdgroup count and raised register pressure, making prefill *slower*. Prefill here is occupancy-bound, not weight-traffic bound.

Cumulative vs the pre-perf reference backend: **prefill ~4.5× (baseline was ~40 tok/s equivalent at this length), decode 2×.** All 25 op parity tests pass, greedy output identical.

digitaloten:
Correction to the last line above: I only measured this commit's delta (107 → 188 tok/s, back-to-back) and the earlier short-prompt cumulative (prefill 8 → 20 tok/s). I did **not** measure the pre-perf reference backend at 614-token prefill, so ignore the "~4.5× / ~40 tok/s" extrapolation — it's not a real data point. Verified claims only: **+76% prefill this commit; 2.5× prefill / 2× decode on the short prompt from the first commit.**

digitaloten:
### Update: profiling-driven kernel parallelization — decode 6.3× on long context

Added a per-op GPU profiler (`INFR_METAL_PROFILE=2`: flushes after each op to attribute GPU commit+wait wall per op) and used it to find where decode actually spends GPU time. Two kernels were running at pathologically low parallelism because decode has `rows=1`:

**1. Norm kernels (`9d310d0`)** — RmsNorm/QkNorm/QkNormRope ran *one thread per normalized group*, so a decode RmsNorm did a whole 1024-dim reduction on a **single** GPU thread. Rewrote as one SIMD group (32 lanes) per group with `simd_sum`. Back-to-back, short context: **GPU 53.6 → 41.8 ms/forward (−22%)**.

**2. Attention (`5916334`)** — one thread per (query, head); for decode that's ~`n_head` threads, each looping the **entire growing KV** serially → attention cost grew with context and dominated long generations. Rewrote as one SIMD group per (query, head): lanes split `head_dim`, score is a `simd_sum` dot, each lane owns a `head_dim/32` slice of the online-softmax accumulator (no cross-lane state, low registers).

| metric (Qwen3-0.6B-Q4_K_M, M3 Pro) | before | after |
|---|---|---|
| decode, 430 tokens | 6.5 tok/s | **41.1 tok/s (6.3×)** |
| GPU/forward @ ~180-tok context | 66.0 ms | **16.0 ms (−76%)** |

Greedy output unchanged, **all 25 op parity tests pass** (`simd_sum` reorders reductions, still within tolerance).

Cumulative decode vs the original reference backend: **3.1 → 41 tok/s (~13×)**, context-dependent. Attention is still the largest remaining GPU consumer (it's inherently O(context)); further gains would need flash-style KV-splitting or Linear/QKV fusion (the latter blocked on the shared cross-backend weight-binding contract — see earlier scope).

digitaloten:
Two more profiling-driven wins pushed (both M3 Pro, Qwen3-0.6B Q4_K_M, 360-token greedy decode, back-to-back runs, parity 25/25 each):

**Split-KV decode attention** (3145c92) — decode has `rows=1`, so the one-simdgroup-per-(query,head) kernel launched only `n_head` (=16) simdgroups and ran O(kv_len) on that fixed tiny width. Each (query,head) now gets a threadgroup of 8 simdgroups, each doing a private online softmax over a strided KV slice, merged in threadgroup memory. Routed only when `rows*n_head` is small — prefill keeps the leaner kernel (the split kernel's static ~8 KB threadgroup memory would cap prefill occupancy).

**Nibble-packed Q4 weights** (ec9d4de) — decode GEMV is bound on the weight byte stream, and the unified form spent a whole u8 per 4-bit code. Q4-family codes now pack two per byte (detected from the data, so exact) and route to a `linear_qui4` kernel that unpacks inline: ~6 bpw resident vs ~10. Reconstruction stays bit-for-bit `infr_gguf::dequant`; Q6_K keeps the u8 kernel.

| metric (360-tok decode) | before | split-KV | +nibble-pack |
|---|---|---|---|
| GPU ms/forward | 21.9 | 11.3 | **9.7** |
| decode tok/s | 44.8 | 85.3 | **98.3** |

ABA check on split-KV: 11.29 → 21.88 (baseline) → 11.31 ms/forward. Nibble-pack repeated twice within 0.1 tok/s. Greedy outputs unchanged ("Paris" + identical ocean paragraph).

Per-op GPU share after both: Linear ~46%, norms ~23%, attention 8%. Attention is no longer the decode bottleneck; remaining headroom is mostly Linear (Q6_K packing, sm stream) and fixed small-op overhead.

digitaloten:
One more (aba1d28): **factored quant weight form**. Half of the Q4_K weight stream was scale metadata — a f32 `(scale, min)` pair per 16 elements. Every affine GGUF quant really has two-level scale structure (integer multipliers × f16 super scales), so the device now keeps exactly that: i16 `(sc, m)` per 16 elems + f16 `(d, dmin)` per quant block (~2.1 bpw vs 4), with codes bit-packed to 4/6/8-bit by the format's max code — Q6_K's 8→6-bit comes free. Q4_K: ~6.1 bpw resident, Q6_K: ~8.1 (was 8/12). Weight traffic per decode forward: ~703 → ~510 MB (−27%).

Still exact: `infr-gguf` grew `dequant_factored` as the single affine decoder and `dequant_unified` is now its expansion — the kernel's `f32(d)·f32(sc)` is the same f32 multiply the reference computed, and the sign/power-of-two factors folded into `m` are exact scalings (Q6_K's `m = −32·sc` is why i16). The llama.cpp-derived dequant tests (94 across gguf/llama/cpu) pin the refactor; metal parity 25/25 with all three kernels covered (Q4_K→quik4, Q6_K→quik6, Q8_0→quik8).

| metric (360-tok decode) | before | after |
|---|---|---|
| GPU ms/forward | 9.98 | **8.40** (−16%) |
| decode tok/s | 96.3 | **112.9** (+17%) |

ABA 8.37 → 9.98 → 8.40; repeats within 0.1 tok/s. Short-prompt prefill also improved 18 → 50 tok/s (same kernels serve prefill; smaller upload + i16/f16 meta). Greedy outputs unchanged.

Session cumulative: decode 44.8 → 112.9 tok/s (2.5×); from the naive impl ~3 → 113 (~36×).

digitaloten:
Round on "kernel-count / barrier overhead" (a2d4ede) — the hypothesis died, but the probes found the real bottleneck:

**Probes** (committed as ignored tests):
- `dispatch_overhead`: per-kernel dispatch+hazard-barrier inside one command buffer = **~1.2 µs**, serialized == independent. 478 kernels/forward ≈ 0.6 ms → op fusion is not a lever. (Confirmed empirically: a fused Add+RmsNorm kernel gained nothing and was reverted.)
- `read_bandwidth_ceiling`: device compute-read roofline ≈ **133 GB/s** (spec 150).
- `gemv_bw`: quik4/quik6 GEMV achieve 92-106 GB/s — decently close to ceiling; tuning attempts (threadgroup width sweep, uint4-wide loads) moved nothing and were reverted.
- Skip-op timing: **attention cost ~1.8 ms/forward at ~200-ctx vs a ~0.2 ms bandwidth floor** — latency-bound on the serial online-softmax chain (~kv_len/NSG dependent steps per simdgroup).

**Fix**: attnsplit kernels are now one macro instantiated at NSG=8 and NSG=32; decode routes to the 32-way split (4× shorter serial chain) when kv_len ≥ 128 and head_dim ≤ 128 (16 KB threadgroup accumulator; hd>128 wouldn't fit 32 KB). New long-context attention parity test exercises the path.

**Measured**: decode 113.8 → 122-127 tok/s (+7-12%), grows with context length. Parity 29/29, greedy output unchanged.

Decode budget now (~8 ms/token): Linear ~4.8 ms (at 92-106 GB/s vs 133 ceiling — the remaining gap is real but two tuning attempts found no purchase), attention ~0.5 ms, small ops ~1.5 ms (169 latency-bound kernels; fusion demonstrably doesn't pay), cb+host ~0.7 ms.

Session cumulative: 44.8 → ~125 tok/s (2.8×); from naive impl ~3 → 125 (~40×).

digitaloten:
Prefill round (25033c6) — **prefill 399 → ~875 tok/s (2.2×)** on an 889-token prompt.

Profiling story: prefill was Linear-bound (77% of GPU). The obvious theory — weights re-streamed once per row by the GEMV kernel — died first: a row-tiled variant (decode each weight block once for 8 rows) measured **zero change**; the SLC was already absorbing the re-reads, and the actual limit is the scalar pipeline's ~1:1 load:FMA structure. So the fix is the matrix pipeline:

- `linear_quik*_mm`: one simdgroup per **8×16 output tile**. Per 16-wide K block, lanes 0-15 decode one output row's block each into a threadgroup tile (same factored-form DEC macros, still bit-exact dequant), then four 8×8×8 `simdgroup_multiply_accumulate`s consume it — every loaded operand reused 8× by hardware. Partial last row-tile takes a scalar fallback; routed at m ≥ 16, out_f % 16 == 0.
- Prefill attention now routes through the split-KV kernels at kv_len ≥ 128 (was a serial O(kv_len) chain per (query,head)) — consistent +6% (4-run A/B).

| 889-tok prompt | before | after |
|---|---|---|
| prefill tok/s | 399 | **~875** |
| Linear GPU ms | 1580 | **456** |

Decode untouched (m=1 path) at ~126 tok/s. Parity 31/31 incl. new m=18 GEMM tests covering full tiles + the partial-tile fallback. Greedy outputs unchanged.

Remaining prefill budget (~1.0s): Linear 456 ms (next: bigger tiles / f16 staging / double-buffering), attention ~340 ms (real fix = flash-style simdgroup_matrix attention — the next big build), small ops ~100 ms.

digitaloten:
Flash-attention round (2b724c5) — negative result, documented and reverted.

Built the full `simdgroup_matrix` flash kernel: 8-query×head tiles, K/V staged f32 to threadgroup memory, scores + P·V as 8×8×8 MMAs, diagonal-MMA row rescale, scalar masked online-softmax phase, partial-tile serial fallback. Parity-clean (27/27 with a new wide-launch test). Iterated three times: direct-device Q fragments, float4/half4 vectorized staging, corr==1 rescale skip.

**It never beat the scalar 32-way split-KV kernel**: 742 → 820 → 842 → 816 tok/s vs 878 (889-token prompt). Diagnosis: occupancy-starved (8 KB threadgroup tile per 32-thread group → ~3 concurrent simdgroups/core) plus a serial stage→barrier→MMA→scalar chain per 8-position block. A chunked-SIMD variant (lane-per-position scoring, 2 reductions per 32 positions) also lost (768) — uncoalesced K loads. Getting MMA attention to win here needs the llama.cpp-style half-precision fragment path and multi-simdgroup tiles; parked.

What shipped: the routing that measured best — split-KV attention for **any** kv_len ≥ 128, prefill included (4-run A/B: 858/889 vs 805/834) — plus the wide-launch parity test the suite lacked, and the negative result recorded in a comment at the routing site so it doesn't get re-attempted blind.

Steady state on Qwen3-0.6B Q4_K_M (M3 Pro): **prefill 885 tok/s** (889-tok prompt, was 399 at session start of this area), **decode 127 tok/s** (~3 at session start). Both within ~30-40% of the measured device ceilings for their bound resource.

digitaloten:
Post-PR validation + llama.cpp scoreboard (same machine, same GGUFs, llama.cpp brew build 4f31eedb, Metal):

| model | shape | llama.cpp | infr-metal | gap |
|---|---|---|---|---|
| Qwen3-0.6B Q4_K_M | pp889 | 4187 tok/s | ~885 | 4.7× |
| Qwen3-0.6B Q4_K_M | tg360 | 192 | ~126 | 1.53× |
| Qwen3-4B Q4_K_M | pp889 | 630 | 134 | 4.7× |
| Qwen3-4B Q4_K_M | tg360 | 46.8 | 32.2 | 1.45× |

**4B validation**: correct greedy output, no shape-routing issues (hd=128, all out_f % 16 == 0 → GEMM path throughout), and decode is exactly at the factored-form bandwidth prediction (~3.1 GB/forward at ~105 GB/s ≈ 32 tok/s) — the backend scales as modeled.

**Gap analysis → next levers, in value order:**
1. **Decode (1.45-1.53×): native-block in-kernel dequant.** llama.cpp reads raw Q4_K blocks (~4.85 bpw); our factored form pays ~6.1-8.1 bpw — the byte ratio (1.37×) explains most of the decode gap. Decoding original GGUF blocks directly in MSL (k4 6-bit scale unpack in-kernel) removes the repack inflation. Est. decode → ~155-165 (0.6B) / ~42 (4B).
2. **Prefill (4.7× at both scales): half-precision-fragment MMA matmul** — the parked flash/GEMM work, now with a measured target. llama.cpp's kernels stage dequantized tiles as half fragments; ours are f32 with narrower tiles.

Both are metal-local follow-ups; neither blocks this PR.

mxaddict (approved):
Reviewed — high quality, merging. Every change profiling-driven and A/B'd, negative results documented in the PR and kept in code comments, evidence tooling committed, and the macOS CI job runs the full parity suite against a real Metal device (green). The only shared-code change (`infr-gguf/dequant.rs`) is a clean refactor: `dequant_factored` as the single affine decoder with `dequant_unified` as its expansion — I checked the bit-identity argument (same f32 multiplies, e.g. Q4_0 `d·1 / d·(−8)` vs the old `d / −8·d`), and the existing llama.cpp-derived tests pin it on Linux CI too.

A few **non-blocking** notes:

1. **`Factored` could serve the Vulkan unified-quant path too.** The Vulkan side's non-native quant path still uploads per-16-block f32 `(scale, min)` pairs (64 bits of metadata per 16 elems where this form uses ~2 bpw). Not urgent — Vulkan mostly runs raw native blocks now — but if we ever revisit the unified path, this is the representation to converge on.

2. **Split-KV decode chunking policy.** Your NSG-simdgroups-per-(query, head) split is the same evolution the Vulkan seam decode just went through (adaptive ~32 chunks/head, 64..512 clamp, and as of today a GPU-side live-count prologue for the record-once replay path — `attn_live.comp`). Worth keeping the two policies conceptually aligned when either changes; a shared comment pointing at each other would do.

3. **Thread-cap gating** (degrading to narrower kernels under low `maxTotalThreadsPerThreadgroup` instead of silently clamping) is exactly right — same class of issue as our Vulkan flash shared-memory-limit gating (`FLASH_SHARED_PER_ROW`). If a third case appears, it may be worth a small shared "capability gate" helper on the Backend trait.

4. **Probe tests as `#[ignore]`d evidence** (dispatch overhead, bandwidth ceilings) is a pattern worth replicating on the Vulkan side — we accumulated similar micro-benches this week (`gemm_bench.rs`) and consistency between backends would be nice.

Thanks — the 3.1 → 126 tok/s decode arc with the dead ends documented is exactly how we want perf work recorded.

PR #5, opened 2026-07-02
perf(metal): native K-quant decode, half-fragment GEMM + flash attention

digitaloten:
**Stacked on #4** — contains its commits until it merges; review only the top commit (04e1370).

The factored quant form paid ~6.1/8.1 bpw where raw GGUF blocks are ~4.5/6.6, plus a host repack (seconds of load time) and a full extra resident copy of every weight. Q4_K and Q6_K now have native kernels decoding the raw block bytes — `weight_qui` binds the bound weight buffer **as-is**. Still bit-for-bit the reference dequant (same f32 `d·sc` products, same `scale·q + min` op order). Other affine formats keep the factored form.

Quant Linear kernels restructured around self-contained per-format `DEC16_*` macros × three shared shapes (GEMV / row-tiled / simdgroup_matrix GEMM). A/B: restructure alone is perf-neutral.

| M3 Pro, back-to-back | factored | native |
|---|---|---|
| 0.6B decode | 126 | **131** |
| 0.6B prefill 889 | 885 | **1079** |
| 4B decode | 32.2 | **37.5** (+16%) |
| 4B prefill 889 | 134 | **180** (+34%) |
| 4B model load | ~2.4 s | **~0.2 s** |
| 4B resident weights | +3 GB copy | raw mmap only |

Divergence lessons kept in comments: Q6_K's 4-way sub-block branch serialized the simdgroup (2.3× slower until branch-free); select-based k4 measured no better than the branch.

Parity 34/34 incl. new m=1 GEMV tests (decode's route was previously untested). llama.cpp decode gap narrows to ~1.25× (4B: 37.5 vs 46.8); remaining delta is access-pattern tuning (their lane mapping), not bytes.

### DIFF
--- a/crates/infr-metal/src/exec.rs
+++ b/crates/infr-metal/src/exec.rs
@@ -396,10 +396,13 @@ impl MetalBackend {
         )
     }
 
-    /// Build (or fetch from cache) a quantized weight in factored device form
-    /// (`dequant_factored`), bit-packing the codes to the narrowest width the format's max code
-    /// fits (4/6/8 bits, low bits first) — decode GEMV is bound on the weight byte stream, so
-    /// every bit shaved here is throughput.
+    /// Build (or fetch from cache) a quantized weight in on-device form. Q4_K and Q6_K — the
+    /// formats real checkpoints ship — have NATIVE kernels that decode the raw GGUF block bytes,
+    /// so the bound weight buffer is used as-is: no host repack, no extra residency, and the
+    /// weight stream stays at the format's true size (~4.5 / ~6.6 bpw vs the factored form's
+    /// ~6.1 / ~8.1). Everything else goes through the factored form (`dequant_factored`), codes
+    /// bit-packed to the narrowest width the format's max code fits. Decode GEMV is bound on the
+    /// weight byte stream, so every bit shaved here is throughput.
     fn weight_qui(
         &self,
         id: TensorId,
@@ -410,6 +413,23 @@ impl MetalBackend {
         let key = buf as *const _ as usize;
         if let Some(w) = self.qui_cache.lock().unwrap().get(&key) {
             return w.clone();
+        }
+        let native_kern = match g.desc(id).dtype {
+            DType::Q4K => Some("linear_q4k"),
+            DType::Q6K => Some("linear_q6k"),
+            _ => None,
+        };
+        if let Some(kern) = native_kern {
+            // The kernel never reads scm/dd for native formats; bind tiny dummies.
+            let w = Arc::new(QuiWeight {
+                codes: buf.raw.clone(),
+                scm: self.zeros_buf(1),
+                dd: self.zeros_buf(1),
+                kern,
+                dshift: 0,
+            });
+            self.qui_cache.lock().unwrap().insert(key, w.clone());
+            return w;
         }
         let bytes = Self::read_bytes(buf);
         let f = infr_gguf::dequant::dequant_factored(g.desc(id).dtype, &bytes);
@@ -594,6 +614,8 @@ impl MetalBackend {
                     let mm_kern = match qw.kern {
                         "linear_quik4" => "linear_quik4_mm",
                         "linear_quik6" => "linear_quik6_mm",
+                        "linear_q4k" => "linear_q4k_mm",
+                        "linear_q6k" => "linear_q6k_mm",
                         _ => "linear_quik8_mm",
                     };
                     let mm_ok = m >= 16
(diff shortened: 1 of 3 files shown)


digitaloten:
Second lever added (3e99a36): **half-fragment prefill GEMM** — f16 operands, f32 accumulate (Apple's mixed-precision simdgroup MMA, the llama.cpp trade), 32×16 output tiles per simdgroup.

Evidence trail: half fragments at the unchanged 8×16 tile measured **zero gain** — the f32 GEMM was staging/barrier-bound, not MMA-rate-bound. The f16 win only cashes as bigger tiles: halved staging + doubled matrix rate → 32×16 (each staged weight block serves 4× rows). 64×16 regressed (register pressure), reverted.

Precision policy made explicit: this is the one path not bit-exact vs the dequant reference (operands round to f16, ~5e-4 relative, well under quant error). Decode GEMV and row-tiled stay exact f32. GEMM parity tests compare against a reference that rounds operands identically, keeping tolerance at 1e-3 (a loosened-tolerance approach was tried first and rejected — synth weights are ~50× real magnitude, it would have tested nothing). Dead f32 GEMM kernels removed.

| 889-tok prefill | PR #4 | +native | +hgemm | llama.cpp |
|---|---|---|---|---|
| 0.6B | 885 | 1079 | **1355** | 4187 |
| 4B | 134 | 180 | **264** | 630 |

Decode unchanged (131 / 37.5 — m=1 never routes here). Parity 34/34. Prefill gap vs llama.cpp: 4.7× → **~2.4-3.1×**; prefill Linear is now 393 ms vs attention's 306 — attention (scalar split-KV) is the next co-dominant target, likely a half-fragment flash retry now that the staging economics changed.

mxaddict (approved):
Reviewed (both the original stack top and the half-fragment GEMM update) — merging. All 5 CI jobs green including the on-device Metal parity suite. Non-blocking notes:

1. **Native raw-block decode is the right convergence point** — this is structurally the same architecture as the Vulkan backend's native path (raw GGUF blocks bound as-is, in-kernel dequant, zero repack/residency). The two backends now agree on the hot-format weight representation, which also supersedes my factored-form note on #4 for Q4_K/Q6_K.

2. **The m=1 GEMV parity tests close a real gap** — decode's kernel route was previously untested. Good catch; we hit the same class of blind spot on Vulkan this week (the replay-attention route had no kernel-level unit until it broke).

3. **The f16-operand GEMM trade is handled honestly**: proving the f32 GEMM was staging-bound before spending the precision (same-tile f16 = zero gain, documented), cashing it as bigger tiles, and adjusting the GEMM parity reference to round operands the same way — keeping the 1e-3 tolerance meaningful rather than just widening it — is exactly right. For the record this matches the Vulkan side (coopmat GEMMs stage f16 operands) and llama.cpp. Decode staying exact f32 is the correct line to hold.

4. **Q6_K branch-free decode lesson**: same finding as Vulkan's Q2_K/Q3_K run-hoisted dequant this week (branchy per-element sub-block decode serializes the SIMD). If a third format needs the treatment, the pattern is now well-established on both backends — worth a shared doc comment someday.

Prefill 1355 t/s on 0.6B / 264 on 4B with the gap to llama.cpp narrowed to access-pattern tuning is a great place for the Metal backend to be.

digitaloten:
Third lever (16de559): **half-fragment flash attention** — the retry of the idea that lost in f32, and this time it wins big.

Why it lost before / wins now: the f32 version spent its time converting K/V through 8 KB threadgroup tiles (choking occupancy) in a stage→barrier→MMA→scalar chain. Half fragments dissolve exactly that: **K^T and V fragments load directly from the f16 KV cache** (strided `simdgroup_load`, transposed for K), Q is cast once per op, threadgroup memory shrinks to the 8×8 score/diag scratch. Accumulation stays f32; P rounds to f16 (same trade as the GEMM). Routed for wide launches on an f16 cache only; decode and small shapes keep the exact scalar kernels.

| 889-tok prefill | PR #4 | +native | +hgemm | +flash | llama.cpp |
|---|---|---|---|---|---|
| 0.6B | 885 | 1079 | 1355 | **~1958** | 4187 |
| 4B | 134 | 180 | 264 | **320** | 630 |

Attention GPU 306 → 105 ms; decode unchanged (131 / 37.5). Parity 30/30 with a dedicated flash test (5e-3 for the f16 Q/P rounding; the wide short-kv test intentionally stays on the exact path).

Prefill gap vs llama.cpp now **~2.0-2.1×**, all of it Linear again (411 ms, 66%): next steps there are staging-free weight fragments (dequant into registers feeding MMAs without the threadgroup round-trip) or wider K-blocks. Cumulative this PR: prefill 2.2× / 2.4×, plus the load-time and memory wins from native blocks.

PR #6, opened 2026-07-02
perf(metal): flash attention + mul_mm-shape cooperative GEMM for prefill

digitaloten:
Follow-up to #5 — this commit was pushed to the branch minutes after the merge, so it needs its own PR. One commit, cherry-picked onto main.

Retry of the flash-attention idea that lost in f32 (documented in #4): that version spent its time converting K/V through 8 KB threadgroup tiles (choking occupancy). Half fragments dissolve exactly those costs — **K^T and V 8×8 fragments load directly from the f16 KV cache** (strided `simdgroup_load`, transposed for K), Q is cast once per op to f16, and threadgroup memory shrinks to the 8×8 score/diag scratch. Scores + output accumulate f32; online softmax stays scalar f32 per 8-position block with the row rescale as a diagonal f32 MMA; P rounds to f16 (same trade as the half-fragment GEMM). Routed only for wide launches on an f16 cache (rows·n_head ≥ 128, kv_len ≥ 64, hd ≤ 128); decode and small shapes keep the exact scalar kernels.

| 889-tok prefill | before | after | llama.cpp |
|---|---|---|---|
| Qwen3-0.6B | 1355 | **~1958** (+44%) | 4187 |
| Qwen3-4B | 264 | **320** (+21%) | 630 |

Attention GPU 306 → 105 ms; decode unchanged (131 / 37.5). Parity 30/30 incl. a dedicated flash test (5e-3 tolerance for the f16 Q/P rounding; the wide short-kv test intentionally stays on the exact path).

### DIFF
--- a/crates/infr-metal/src/exec.rs
+++ b/crates/infr-metal/src/exec.rs
@@ -908,7 +908,12 @@ impl MetalBackend {
                 // beat the 32-way split on prefill — occupancy-starved by its threadgroup tiles —
                 // so it was dropped.)
                 let f16 = g.desc(k_cache).dtype == DType::F16;
-                let split = rows * nh < 128 || kv_len >= 128;
+                // Wide launches on an f16 cache take the half-fragment flash kernel: K^T/V
+                // fragments load straight from the cache, Q is cast once to f16 below. Small
+                // kv_len stays scalar (also keeps the short-kv wide parity test on the exact
+                // path). See `attnflash_f16kv` for why the f32 flash attempt lost and this wins.
+                let flash = f16 && rows * nh >= 128 && kv_len >= 64 && hd <= 128 && hd % 8 == 0;
+                let split = !flash && (rows * nh < 128 || kv_len >= 128);
                 // The split kernels REQUIRE their full NSG*32-thread threadgroup: every simdgroup
                 // owns a strided KV slice, so a smaller launch would silently skip positions and
                 // merge uninitialized partials. maxTotalThreadsPerThreadgroup is per-PIPELINE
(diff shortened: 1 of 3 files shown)


mxaddict (approved):
Reviewed — merging. Non-blocking notes:

1. **This is the model example of a negative result paying off**: the f32 flash loss in #4 was documented with its mechanism (threadgroup staging choke), so this retry could target exactly that — fragments loading straight from the f16 KV cache. Keeping dead ends recorded with a cause is what made the +44% possible; worth preserving that discipline as the backend matures.

2. **Routing gates are conservative and right** (rows·nh ≥ 128, kv ≥ 64, hd ≤ 128, f16 cache; decode + small shapes stay exact). One thought for later: hd ≤ 128 excludes gemma-class models (hd=256) — the Vulkan side hit the same wall and extended its general paths to hd ≤ 512 this week. Not needed until Metal targets gemma.

3. **Precision line holds**: f16 operands / f32 accumulate for throughput paths, exact f32 for decode — consistent across #5 and #6, matching the Vulkan coopmat trade. The dedicated flash parity test at a justified 5e-3 plus deliberately keeping the wide short-kv test on the exact path (so strict coverage doesn't erode) is exactly how to manage this.

Attention 306 → 105 ms with decode untouched — the Metal prefill story (399 → ~1958 t/s across the three PRs) is a great arc.

digitaloten:
Second commit (bae5910): **cooperative-tile GEMM** — one 32×64 tile per 128-thread threadgroup instead of private 32×16 tiles per simdgroup. Threads 0-63 decode the shared weight tile while threads 64-127 stage the x tile concurrently; all four simdgroups then consume the shared x (device traffic /4).

Evidence: the private-tile GEMM measured ~3.1 TFLOPS vs ~14 peak; probes ruled out barrier frequency and decode width (a private-tile BK=32 variant measured nothing, a cooperative BK=32 variant measured worse — both documented, neither kept).

| 889-tok prefill | +flash | +coop | llama.cpp |
|---|---|---|---|
| 0.6B | 1971 | **~2114** | 4191 |
| 4B | 321 | **358** | 630 |

Decode unchanged. Parity 32/32 with dedicated cooperative-tile tests (full + partial 32-row tiles).

Honest status: prefill gap now ~1.8-2.0×, all Linear. Three tile-shape experiments in, gains are flattening at ~10%/step; closing the rest likely requires the full mul_mm treatment (larger BM with different accumulator distribution, operand swizzling). Diminishing returns from here — good stopping point for this PR.

digitaloten:
Third commit (400d521): the **full mul_mm treatment** — and it delivered what the incremental tile experiments couldn't.

Ported llama.cpp's kernel_mul_mm shape onto our DEC16 quant decoders. The pieces that turned out to be the actual gap (each previously probed around, none sufficient alone):
- **Contiguous 8×8 half tiles in threadgroup memory** (weights written pre-transposed): all simdgroup_loads become stride-8 and conflict-free — this, not barrier count or decode width, was the missing structure.
- 64×32 tile per threadgroup, NK=32, one 16-block dequant per thread per K-step, x cast f32→f16 **inline during staging** (separate cast pass deleted).
- Partial token tiles stage through shared memory with row-guarded copy-out — the scalar tail path is gone; everything MMAs.

| 889-tok prefill | +coop | +mul_mm | llama.cpp | gap |
|---|---|---|---|---|
| 0.6B | 2114 | **~2840** | 4191 | 1.48× |
| 4B | 358 | **~498** | 630 | **1.27×** |

Decode unchanged (131.7 / 37.5). Parity 32/32; greedy outputs unchanged.

PR cumulative: 0.6B prefill 1355 → 2840 (2.1×), 4B 264 → 498 (1.9×), on top of #4/#5. Prefill at 4B is now within 27% of llama.cpp; remaining delta is spread thin (their per-format dequant micro-tuning, attention at 0.6B scale). Clean stopping point.

PR #7, opened 2026-07-02
docs(metal): architecture doc + CONTEXT/README refresh

digitaloten:
Documents the Metal backend as it now exists after #4/#5/#6:

- **`docs/METAL.md`** (new): execution model (residency, one command buffer per forward) with its measured floors; the two quantized weight forms and their exactness contracts; the `DEC16_*` decoder × kernel-shape matrix with per-path precision policy (decode bit-exact f32, prefill GEMM f16 operands/f32 accumulate); attention routing including the per-pipeline thread-cap gating that the paravirtual CI device forced; profiler + probe tests; the full negative-results list so dead ends don't get re-attempted; remaining levers with estimates.
- **CONTEXT.md**: Metal section with current numbers vs llama.cpp (same GGUF/machine), the load-bearing lessons, and the bench invocation (metal isn't wired into `infr bench` yet — noted as an open TODO).
- **README**: Metal line updated from "correctness-first reference" to the optimized state, linking the new doc.

No code changes.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -35,6 +35,39 @@ decisions: the `compute-backend-direction` memory.** Short version:
   adapter (`compile`/`execute`, still `todo!()`) → gemma4/E2B/MoE on the seam →
   cleanup.
 
+## Metal backend (Apple Silicon) — REBUILT FOR PERF (2026-07-02, PRs #4/#5/#6)
+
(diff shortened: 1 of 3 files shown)


mxaddict (approved):
Docs match the code as merged through #6 — the negative-results list and the per-path precision policy table are exactly what future sessions need to not re-attempt dead ends. Merging.

PR #8, opened 2026-07-02
perf(metal): mul_mm GEMM recovery, mul_mv decode GEMV, 16-wide flash

digitaloten:
Four commits. The first two are **recovery**: #6 merged minutes before the cooperative-tile and mul_mm GEMM pushes landed on its branch, so those commits never reached main (same race as #5 → #6) — cherry-picked here unchanged. The last two are new work.

1. `bae5910` cooperative-tile GEMM (32×64 per threadgroup, shared staging) — prefill +11%/+7%
2. `400d521` **mul_mm-shape GEMM** (contiguous pre-transposed 8×8 half tiles in threadgroup memory, NK=32, inline f32→f16 x cast, MMA'd partial tiles) — prefill +39%/+34%; the smem tile layout, not tile size or barriers, was the gap
3. mul_mv-shape decode GEMV for native Q4_K/Q6_K (2 rows/simdgroup, activations register-reused across rows, decode-free masked-nibble inner loop with per-group scales; algebraically identical, FP-reassociated) — **decode +7.6%/+6.4%**
4. flash attention: 16 KV positions per iteration (one scalar softmax phase + rescale per 16, half the barriers) — 0.6B prefill +6%

| M3 Pro, 889-tok | main today | this PR | llama.cpp | gap |
|---|---|---|---|---|
| 0.6B decode | 131.5 | **143.8** | 191.4 | 1.33× |
| 0.6B prefill | ~1970 | **~3019** | 4191 | 1.39× |
| 4B decode | 37.5 | **39.9** | 46.7 | **1.17×** |
| 4B prefill | ~321 | **~519** | 630 | **1.21×** |

Parity 30/30 throughout; greedy outputs unchanged. Suggest holding the merge until the head is confirmed final — the last two PRs merged mid-push.

PR #9, opened 2026-07-02
feat(bench): metal in infr bench + long-context depth sweep findings

digitaloten:
Stacked on #8 (contains its commits until it merges; review the top commit).

**Metal in `infr bench`** (`INFR_METAL=1` or `--dev Metal`): same pp/tg/pg + depth methodology as the CPU arm through the agnostic seam. Metal numbers are now llama-bench-comparable; sanity-checked against the run-path figures.

**First depth sweep** (Qwen3-0.6B, M3 Pro, vs llama.cpp Metal, same GGUF):

| shape | infr | llama.cpp | gap |
|---|---|---|---|
| tg128 @ d4096 | 68.0 | 98.8 | 1.45× |
| tg128 @ d16384 | 25.5 | 46.1 | 1.81× |
| tg128 @ d32768 | 13.9 | 23.4 | 1.68× |
| pp8192 | 785 → **875** | 2336 | 2.7× |

Long context — the north star — is the weakest axis. Profiling: prefill attention = **80% of pp8192 GPU** (7.5 s vs Linear's 1.6 s).

**One fix landed from the evidence**: the flash grid ran heads fastest, so concurrent query tiles streamed 16 per-head KV regions (~64 MB) and thrashed the SLC; same-head tiles are now adjacent (+11% pp8k, short prefill unchanged, parity 32/32).

**Identified next arc**: port `flash_attn_ext`'s structure (KV split across simdgroups within the tile, half-precision S/P, deeper pipelining) — same FLOPs, ~4× faster at depth. Decode at depth points at the same class (attnsplit32's serial chains).

### DIFF
--- a/crates/infr-cli/src/main.rs
+++ b/crates/infr-cli/src/main.rs
@@ -752,6 +752,21 @@ fn cmd_bench(
             json,
         );
     }
+    // INFR_METAL=1 (or --dev Metal): bench the dense forward on the Metal backend through the
+    // agnostic seam — same pp/tg/pg + depth methodology as the CPU arm, directly comparable to
+    // `llama-bench` on the Metal build.
+    if std::env::var("INFR_METAL").is_ok() || dev.eq_ignore_ascii_case("metal") {
+        return cmd_bench_metal(
+            &gguf,
+            tok.as_deref(),
+            n_prompt,
+            n_gen,
+            depth,
+            pg,
+            reps,
+            json,
+        );
+    }
     let _ = dev; // GPU device selection: VulkanBackend uses the default adapter (--dev reserved for parity).
     let llama = infr_llama::Llama::load_opt(&gguf, tok.as_deref())?;
     let measure_tg = pg.is_none() && n_gen > 0;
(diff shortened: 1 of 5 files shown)


mxaddict (commented):
Reviewed the full stack (the four #8 commits ride along here, so this covers both PRs). Content is approve-worthy — evidence discipline is again exactly right (the 3.1-vs-14-TFLOPS diagnosis with BK-width and barriers ruled out by probes before porting mul_mm; the 12ms-of-105ms KV-bandwidth measurement justifying the 16-wide flash). Three items before merge:

1. **CI clippy is a merge artifact against today's main, not your code**: main's bench-reporter refactor (95e480d..) moved `cmd_bench_cpu`'s doc + `#[allow(clippy::too_many_arguments)]`, and this branch re-adds the allow at the old spot → `duplicated attribute` at main.rs:891 and an un-allowed 8-arg fn at :939 on the merge commit. A rebase over main resolves both; `cmd_bench_metal` can also use the shared `print_bench_avg` reporter that refactor introduced (your tail block is a third copy of the pattern it deduplicated).

2. **Doc drift vs #7 (non-blocking but worth a one-liner)**: the mul_mv decode GEMV commit describes itself as "algebraically identical, FP-reassociated" — which quietly retires METAL.md's "decode bit-exact f32" contract for Q4_K/Q6_K. The parity tolerance covers it and the trade is right; just update the exactness table in METAL.md so the doc doesn't promise more than the code.

3. **Parity coverage note (fine as-is)**: the mul_mm and mul_mv commits add no new parity cases because they swap kernels behind ALREADY-covered routes (the m=1 GEMV tests from #5 exist precisely for this) — good; the +32 lines on the cooperative-tile commit cover the one new route.

On the process point from #8: happy to merge #9 alone once rebased (it strictly contains #8, which then closes as superseded) — just confirm the head is final first, given the three post-merge orphan races this week.

mxaddict:
Housekeeping: closed #8 as superseded by this PR — all four of its commits are contained here with identical SHAs, so merging this lands the full stack. Its hold-until-head-final request now applies here.

mxaddict:
Closing as superseded: #10 contains all of this PR's commits with identical SHAs (the full #8 stack + cf2d5b4) plus the flash_attn_ext work on top. The open review items here (clippy merge-artifact rebase over main's bench refactor, METAL.md 'decode bit-exact' doc drift) carry over to #10.

PR #10, opened 2026-07-02
perf(metal): flash_attn_ext ports — long-context prefill 2.5x, decode at depth 1.9x

digitaloten:
Ports the two llama.cpp flash-attention kernel structures that dominate long-context cost. **Stacked on #9** (contains #8 + #9's commits; the last four commits are new — merge those first, or merge this one to take the whole stack). The depth sweep in #9 identified attention as 80% of pp8192 GPU time.

## Cooperative flash attention (prefill): `attnflash2_f16kv`

The `kernel_flash_attn_ext` structure — NSG=4 simdgroups cooperate on ONE (8-query tile, head) threadgroup over C=64 KV positions per iteration, with each phase split along a different axis so every lane stays busy (the previous single-simdgroup kernel stalled in its scalar softmax, 8 of 32 lanes active, one phase per 16 KV):

- **QK^T** — the 8 score fragments split across simdgroups; K^T fragments load directly from the f16 cache (transposed, no staging).
- **softmax** — split by query rows, one float2 per lane over the 64 scores, all 32 lanes active; each row's online max/sum stats stay pinned in the owning simdgroup's registers for the whole KV loop.
- **P·V** — split by output columns, O fragments held in registers; paired KV blocks keep 2 P + 2·no V loads in flight.

Masking is analytic (causal + sliding window per row) — no mask buffer. Partial query tiles zero-pad Q in shared memory (drops the serial fallback path). Templated over compile-time head_dim, instantiated hd=64/128 (fully unrolled loops); NSG=8 was benched and lost. Pipeline-cap gated at 128 threads; other head sizes keep the old flash kernel.

## Vector flash attention (decode): `attnvec_f16kv`

The `kernel_flash_attn_ext_vec` structure for long-context decode — same split-KV idea as `attnsplit32`, but each simdgroup step covers **32 KV positions instead of 1**: lanes fold as 4 KV rows × 8-lane dots with a shuffle-tree reduce, one simd_max/simd_sum softmax pass per block, log2 merge of per-simdgroup partials at the end. The split kernels were latency-bound on their per-position serial chain. Q stays f32 in shared — same numeric class as attnsplit32 (parity tol 1e-4, no rounding). Tail rows clamp their pointer into the cache and mask in the softmax. attnsplit32 retained for other head sizes and cap degradation.

## Numbers (M3 Pro, Q4_K_M, vs llama.cpp Metal on the same GGUF)

| metric | before | after | llama.cpp | gap |
|---|---|---|---|---|
| 0.6B pp8192 | 864 | **2167** | 2336 | 1.08x |
| 0.6B pp2048 | 1327 | **3346** | 3748 | 1.12x |
| 4B pp8192 | 278 | **422** | 488 | 1.16x |
| 0.6B tg128@d16384 | 25.5 | **48.3** | 42.9 | **infr wins** |
| 0.6B tg128@d4096 | 66.9 | **86.3** | 114.5 | 1.33x* |
| 4B tg128@d16384 | 14.9 | **21.4** | 24.5 | 1.14x |

\* remaining tg gap at moderate depth is per-token engine overhead (graph recompile per token — the parked decode-replay work), not the attention kernel.

Short-context decode and prefill unchanged. Parity suite grown to 37 GPU tests: flash2 at hd=64/128 + sliding window + partial tile, vec at depth + mid-block window edge + tail clamp, hd=72/96 shapes keep the old flash/split32 paths covered. E2E chat verified coherent on metal.

### DIFF
--- a/crates/infr-cli/src/main.rs
+++ b/crates/infr-cli/src/main.rs
@@ -858,6 +858,21 @@ fn cmd_bench(
             json,
         );
     }
+    // INFR_METAL=1 (or --dev Metal): bench the dense forward on the Metal backend through the
+    // agnostic seam — same pp/tg/pg + depth methodology as the CPU arm, directly comparable to
+    // `llama-bench` on the Metal build.
+    if std::env::var("INFR_METAL").is_ok() || dev.eq_ignore_ascii_case("metal") {
+        return cmd_bench_metal(
+            &gguf,
+            tok.as_deref(),
+            n_prompt,
+            n_gen,
+            depth,
+            pg,
+            reps,
+            json,
+        );
+    }
     let _ = dev; // GPU device selection: VulkanBackend uses the default adapter (--dev reserved for parity).
     let llama = infr_llama::Llama::load_opt(&gguf, tok.as_deref())?;
     let measure_tg = pg.is_none() && n_gen > 0;
@@ -976,6 +991,57 @@ fn print_bench_avg(samples: &[f64], label: &str, depth: usize, tag: &str, reps: 
 /// CPU-backend bench (`infr bench -ngl 0`): the GPU bench's pp/tg/pg metrics on the agnostic CPU
 /// reference path, using `CpuModel`'s token-level timing — directly comparable to `llama-bench -ngl 0`.
 #[allow(clippy::too_many_arguments)]
+/// Metal twin of [`cmd_bench_cpu`]: same pp/tg/pg + depth methodology on the Apple-GPU seam
+/// backend (`CpuModel::bench_metal`). On non-macOS this arm is unreachable (the backend crate
+/// compiles to nothing), so the whole body is cfg-gated.
+#[allow(clippy::too_many_arguments)]
+fn cmd_bench_metal(
+    gguf: &Path,
+    tok: Option<&Path>,
+    n_prompt: usize,
+    n_gen: usize,
+    depth: usize,
+    pg: Option<(usize, usize)>,
+    reps: usize,
+    json: bool,
+) -> anyhow::Result<()> {
+    #[cfg(not(target_os = "macos"))]
+    {
+        let _ = (gguf, tok, n_prompt, n_gen, depth, pg, reps, json);
+        anyhow::bail!("metal bench requires macOS");
+    }
+    #[cfg(target_os = "macos")]
+    {
+        let model = infr_llama::CpuModel::load(gguf, tok)?;
+        let measure_tg = pg.is_none() && n_gen > 0;
+        // One untimed warmup: page-cache the mmap + build the weight caches + compile pipelines.
+        let _ = model.bench_metal(depth.max(1), if measure_tg || pg.is_some() { 1 } else { 0 });
+        let mut samples = Vec::with_capacity(reps);
+        for _ in 0..reps {
+            let ts = if let Some((p, g)) = pg {
+                let s = model.bench_metal(p, g)?;
+                (p + g) as f64 / (s.prompt_secs + s.decode_secs)
+            } else if measure_tg {
+                let s = model.bench_metal(depth.max(1), n_gen)?;
+                n_gen as f64 / s.decode_secs
+            } else {
+                let s = model.bench_metal(n_prompt, 0)?;
+                n_prompt as f64 / s.prompt_secs
+            };
+            samples.push(ts);
+        }
+        let label = if let Some((p, g)) = pg {
+            format!("pg{p}+{g}")
+        } else if measure_tg {
+            format!("tg{n_gen}")
+        } else {
+            format!("pp{n_prompt}")
+        };
+        print_bench_avg(&samples, &label, depth, " [metal]", reps, json);
+        Ok(())
+    }
+}
+
 fn cmd_bench_cpu(
     gguf: &Path,
     tok: Option<&Path>,
--- a/crates/infr-cli/src/main.rs
+++ b/crates/infr-cli/src/main.rs
@@ -988,9 +988,6 @@ fn print_bench_avg(samples: &[f64], label: &str, depth: usize, tag: &str, reps: 
     }
 }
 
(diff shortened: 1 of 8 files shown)


mxaddict:
Consolidating the stack: #8 and #9 are closed as superseded — this PR contains their commits with identical SHAs, so **#10 is now the single merge vehicle for the whole Metal arc** (mul_mm/mul_mv/16-wide flash → bench + depth sweep → flash_attn_ext ports). One merge, no more post-merge orphan races.

Carrying over the open items from the #9 review:

1. **Rebase over current main** before merge — main's bench-reporter refactor moved `cmd_bench_cpu`'s `#[allow(clippy::too_many_arguments)]`, and the branch re-adds it → `duplicated attribute` + an un-allowed 8-arg fn on the merge commit (the #9 CI clippy failure). While rebasing, `cmd_bench_metal`'s tail can use the shared `print_bench_avg` reporter main introduced.
2. **METAL.md doc drift**: the mul_mv decode GEMV is "algebraically identical, FP-reassociated", which retires the "decode bit-exact f32" contract merged in #7 — one-line table update.
3. When the head is final, say the word and we merge this one PR.

digitaloten:
Review items addressed, plus two more decode commits stacked here per the consolidation:

**Review items**
1. Rebased over current main — the duplicated `too_many_arguments` attribute and orphaned doc comment are fixed (`cmd_bench_cpu` got its doc/allow back); `cmd_bench_metal` already used the shared `print_bench_avg` reporter after the rebase.
2. METAL.md decode contract updated to 'algebraically identical, FP-reassociated' — plus the attention-routing section now covers the flash_attn_ext ports and the scoreboard reflects the current bench numbers.

**New commits**
- `feat(metal): record-once decode replay on the seam` — implements `Capabilities::decode_replay` (the Vulkan contract): an eligible decode graph records once as a flat dispatch tape, re-encoded per token with dynamic-pos kernels reading the bound positions buffer. Token streams verified identical replay on/off; per-token host cost ~10x down (decode was already GPU-bound, so throughput moves little — the value is the seam contract + headroom).
- `perf(metal): fuse decode Linear+Add residuals` — the metal twin of the Vulkan adapter's `linear_add_peephole`. +2.5% on 4B decode.

Current numbers (0.6B/4B vs llama.cpp same GGUF): tg128 159/40.9 vs 191/46.7; tg128@d16384 48.3/21.4 vs 42.9/24.5; pp8192 2167/422 vs 2336/488. Parity suite at 39 GPU tests, clippy clean.

Not calling the head final just yet — will confirm explicitly when the arc wraps.

digitaloten:
cmm arc landed — two more commits closing the prefill Linear gap (85% of 4B prefill GPU time):

- **GEMM pipelining**: the weight-block dequant + activation loads now issue BEFORE the barrier that drains the previous iteration's MMA phase (the reference mul_mm does this; the port had them serialized), B staging is vectorized (float4→half4 pairs instead of 8 scalar round-trips), and the fragment loads get `simdgroup_barrier(mem_none)` scheduling hints. 4B pp2048 514→543.
- **Reference-shape K-quant decode** in the GEMM/RT kernels: Q4_K keeps the high nibble in place with the /16 folded into the scale (mask instead of per-element shift/select); Q6_K unpacks via uint32 ql/qh lane combines (8 ushort loads instead of 32 byte loads). Bit-identical values, parity suite green unchanged. 4B pp2048 543→576, 0.6B pp2048 3414→3543.

Prefill is now within ~4-6% of llama.cpp at pp2048/pp8192 on both models (0.6B pp8192: 2258 vs 2336). Scoreboard in METAL.md updated. 39 parity tests, clippy clean.

digitaloten:
Head is final at `558f06a` — good to merge.

Final state: 15 commits — mul_mm GEMM recovery + mul_mv decode GEMV + 16-wide flash → metal bench wiring + depth-sweep grid fix → flash_attn_ext ports (cooperative prefill + vector decode) → decode replay on the seam → Linear+Add residual fusion → cmm pipelining + reference-shape K-quant decode. 39 GPU parity tests green, clippy clean, rebased over main with both review items addressed.

Idle-machine scoreboard vs llama.cpp (same GGUF, back-to-back, in METAL.md): prefill within 5-7% on both models, 0.6B decode at d16384 ahead (+7%), 4B decode at depth within 12%.

digitaloten:
Heads-up: the declared-final head failed CI's `cargo fmt --check` (the branch predates a local toolchain update — CI's stable rustfmt is newer than what formatted those commits). Pushed `4e74403`, a pure `cargo fmt` commit, zero semantic changes (2 files, 10 insertions / 11 deletions of whitespace-level reflow). Clippy and the 39-test GPU parity suite re-verified on the new head. **Final head is now `4e74403`** — apologies for moving it after calling it final; CI-green won.

digitaloten:
Head moved again — three housekeeping commits, now `54279c2`:

1. Rebased over current main (the seam MoE/E2B batching + gemma replay work).
2. `fix(metal): honor the widened seam replay contract on untaped decode graphs` — main's runner now replays gemma-family and MoE decode graphs on any `decode_replay` backend, but the metal tape only covers the qwen3-dense shape; every other decode graph would have re-executed with the baked pos=0. The normal path now reads the current position off the bound positions buffer at encode time and overrides baked pos/kv_len in Attention/WriteKv (identity for single-execute callers). Verified qwen3moe decode under the replay loop produces token streams identical to `INFR_SEAM_NO_REPLAY`.
3. `style:` clippy 1.96 lint in the profiler.

39 GPU parity tests + fmt/clippy green on the new head. This is the final head unless main moves under it again.

mxaddict:
Merged manually at 2bee07c — main had moved under you a third time (our GatedActFused/combined_gu Metal edits), so I resolved it on our side rather than ask for another rebase. Resolution: your decode_replay: true + our combined_gu: false in the Capabilities initializer; your replay_shape gate already routes the new ops to the untaped path, so the two changes compose cleanly. Full Vulkan/CPU suites + seam benches verified unchanged before push. Thanks for tracking the seam contract change in 45e1287 — good catch on the untaped pos override. #11 can rebase down to its single MoE commit now.

PR #11, opened 2026-07-02
feat(metal): device MoE path — expert-batched GEMVs off a GPU expert table

digitaloten:
**Draft, stacked on #10** — contains all of #10's commits plus one new commit (`e9e7d6b`); the diff collapses to just the MoE work once #10 merges.

## What

`Op::MoeFfn` previously ran on the host: per token per layer it dequantized three expert weight slices to f32 on the CPU and round-tripped each matvec through its own command buffer. For K-quant experts (the shapes real MoE checkpoints ship) the whole FFN now runs on-device in **seven dispatches**:

1. **Router GEMV** (dequant-cached f32 weight) → **`moe_topk`**: softmax + top-k on device — ties go to the lowest index, matching the CPU reference's stable sort — with renormalized weights written to a 32-slot device expert table.
2. **Gate/up/down GEMVs batched over the selected experts**: one dispatch covers all n_used slots (slot = high grid bits), each resolving its expert's weight slice from the table in-kernel. The shared `linear_q4k`/`linear_q6k` bodies gained an EPI-2 epilogue (`dst = w·s`) so the down projection folds the routing weight per slot.
3. **Gated activation** over all slots at once, and a final **`moe_reduce`** doing the slot-ascending weighted sum (the CPU accumulation order).

Expert scratch lives in a reusable (size, tag)-keyed pool — no per-op allocations. The host path is retained for non-K-quant dtypes and as the `INFR_METAL_NOMOE` escape hatch.

## Evidence

`tests/moe_bw.rs` (Qwen3-30B-A3B layer shape — ne=2048, 128 experts, 8 used, n_ff_exp=768 — M3 Pro, marginal cost in an 8-layer chain):

| path | ms / MoE layer |
|---|---|
| host (previous) | 114.9 |
| device, per-slot dispatches | 0.570 |
| device, expert-batched | **0.465** |

~250× over the host path. Parity: q4k + q6k device-path tests vs the CPU oracle, plus the existing f32 test keeping the host fallback covered (41 GPU tests, clippy clean).

## Caveats / draft status

- **Not yet validated on a real checkpoint** — no qwen3moe GGUF fits this machine's 18 GB (30B-A3B Q4_K_M is 18.6 GB). The op-level parity + probe are the evidence so far; happy to keep this as draft until someone runs a real model, or land it as strictly-better-than-host.
- Known follow-ups: `moe_topk` is serial on one thread (~tens of µs/layer, parallelizable), and prefill currently hits this op per token — a token-batched expert pass is the phase-2 item.

### DIFF
--- a/crates/infr-cli/src/main.rs
+++ b/crates/infr-cli/src/main.rs
@@ -340,11 +340,20 @@ fn cmd_run(model: &str, message: Option<&str>) -> anyhow::Result<()> {
             Box::new(infr_llama::model::Qwen35Chat::new_metal(gguf.clone()))
         } else {
             eprintln!(
-                "[metal backend — dense/MoE forward on Apple GPU via the agnostic compute graph (reference)]"
+                "[metal backend — dense/MoE forward on Apple GPU via the agnostic compute graph, persistent KV session]"
             );
-            Box::new(infr_llama::model::CpuDenseChat::new_metal(
-                infr_llama::CpuModel::load(&gguf, tok.as_deref())?,
-            ))
+            #[cfg(target_os = "macos")]
+            {
+                Box::new(infr_llama::model::MetalSeamChat::new(
+                    infr_llama::CpuModel::load(&gguf, tok.as_deref())?,
+                ))
+            }
+            #[cfg(not(target_os = "macos"))]
+            {
+                Box::new(infr_llama::model::CpuDenseChat::new_metal(
+                    infr_llama::CpuModel::load(&gguf, tok.as_deref())?,
+                ))
+            }
         }
     } else if std::env::var("INFR_CPU").is_ok() {
         if is_q35 {
@@ -1418,9 +1427,18 @@ fn cmd_serve(model: &str, addr: &str) -> anyhow::Result<()> {
         };
         Some(Box::new(m))
     } else if std::env::var("INFR_METAL").is_ok() {
-        Some(Box::new(infr_llama::model::CpuDenseChat::new_metal(
-            infr_llama::CpuModel::load(&gguf, tok.as_deref())?,
-        )))
+        #[cfg(target_os = "macos")]
+        {
+            Some(Box::new(infr_llama::model::MetalSeamChat::new(
+                infr_llama::CpuModel::load(&gguf, tok.as_deref())?,
+            )))
+        }
+        #[cfg(not(target_os = "macos"))]
+        {
+            Some(Box::new(infr_llama::model::CpuDenseChat::new_metal(
+                infr_llama::CpuModel::load(&gguf, tok.as_deref())?,
+            )))
+        }
     } else if std::env::var("INFR_CPU").is_ok() {
         Some(Box::new(infr_llama::model::CpuDenseChat::new(
             infr_llama::CpuModel::load(&gguf, tok.as_deref())?,
--- a/crates/infr-cpu/src/lib.rs
+++ b/crates/infr-cpu/src/lib.rs
@@ -2237,6 +2237,22 @@ fn cpu_buf(b: &dyn Buffer) -> &CpuBuffer {
         .expect("cpu backend: buffer is not a CpuBuffer (mixed backends?)")
 }
 
+/// Dequantize the first `need` elements of a Q8_0-block buffer (34 B / 32 elems, y = d*q).
+pub(crate) fn dequant_prefix_q8_0(bytes: &[u8], need: usize) -> Vec<f32> {
+    let mut out = Vec::with_capacity(need);
+    for b in 0..need.div_ceil(32) {
+        let off = b * 34;
+        let d = half::f16::from_le_bytes([bytes[off], bytes[off + 1]]).to_f32();
+        for i in 0..32 {
+            if out.len() == need {
+                break;
+            }
+            out.push(d * (bytes[off + 2 + i] as i8) as f32);
+        }
+    }
+    out
+}
+
 impl Backend for CpuBackend {
     fn name(&self) -> &str {
         "cpu"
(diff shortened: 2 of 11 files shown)


digitaloten:
Two updates — the draft caveats are now resolved:

**Real-model validation** (found a checkpoint that fits: mradermacher/Qwen3-MOE-4x0.6B-2.4B Q4_K_M, real qwen3moe arch):
- decode: **0.6 → 112 t/s** vs the host path (llama.cpp: 141 t/s on the same GGUF — a 1.27× gap, in line with the dense models)
- prefill: 1 → 105 t/s (still per-token; engine-side batched MoE prefill is the remaining follow-up — the seam builder's per-token guard, not this backend)
- device/host **token streams are bit-identical**

**Lane-parallel `moe_topk`**: the serial one-thread selection loop turned out to be ~half the MoE layer wall. Now each lane offers its best untaken expert from its stride, simd_max/simd_min resolve the winner with the reference's exact stable-sort tie order, and psum/wsum keep the reference's serial summation order (bit-identical results). Probe: **0.465 → 0.224 ms** per 30B-A3B-shaped layer — 95 GB/s effective expert stream, near the 107 GB/s GEMV ceiling.

Happy to mark ready-for-review whenever #10 lands; note this branch contains #10's commits, so the diff collapses on merge.

digitaloten:
Batched MoE prefill added (this was the 25× prefill gap):

- `Op::MoeFfn` carries `rows` (each row routed independently); the seam builder batch-prefills MoE prompts when the backend advertises the new `Capabilities::moe_batched`.
- Metal device path grids the expert stages over (row, slot) with per-row top-k into a [rows, 32] table, chunked at 256 rows to bound scratch. CPU reference loops rows and also sets the capability. **Vulkan untouched**: its single-token id-native path leaves the capability unset, so it keeps the per-token prefill loop exactly as today (the arm rejects rows>1 defensively).
- Also fixes the `pp: inf` bench stat on MoE models (prompt_secs was never set on the per-token prefill path).

Qwen3-MOE-4x0.6B Q4_K_M: **pp512 105 → 605 t/s**; decode unchanged at 111; e2e coherent. llama.cpp is 2651 on the same GGUF — the remaining 4.4× is GEMV-shaped expert compute, and an expert-grouped GEMM pass (mul_mm_id-style) is the named follow-up rather than part of this PR.

fmt/clippy(-D warnings)/tests all green locally against the CI matrix.

digitaloten:
Phase B landed — expert-grouped GEMM for the batched prefill (the mul_mm_id shape):

`moe_map` groups each 256-row chunk's (token, slot) pairs by expert on-device (deterministic, token-ascending per expert), then `linear_*_cmm_id` runs the shared cooperative-GEMM tile over each expert's token group — grid = expert × token-tile × out-tile with early-return past each expert's count, token rows resolved through the id list during staging, output scattered through threadgroup staging with the routing weight folded into the down projection's scatter. Decode (rows < 16) keeps the GEMV stages.

MoE prefill now rides the same f16-operand / f32-accumulate GEMM policy as dense prefill, and the batched parity test mirrors that rounding in a hand-rolled reference.

**Qwen3-MOE-4x0.6B Q4_K_M progression on this PR: pp512 105 → 605 → 2013 t/s** (llama.cpp: 2651 on the same GGUF — 25× behind at the start of this arc, 1.32× now). Decode steady at ~113 t/s. fmt/clippy/tests green locally; CI still isn't spawning check suites for this PR (GitHub-side, needs a poke from your end when convenient).

digitaloten:
Two updates:

**Rebased over current main** — and integrated with the seam MoE work that landed there: dropped my `Op::MoeFfn.rows` field and `Capabilities::moe_batched` in favor of main's conventions (rows inferred from the bound activation shape, the runner's dtype-gated batching). The metal backend now just executes what the seam already emits; Vulkan/engine untouched by this PR. Also picked up +5% MoE prefill from main's seam improvements (pp512 now 2126).

**Persistent KV session on the Metal seam** — the Metal twin of the Vulkan session: `DenseMetalSession` owns the backend/weights/KV across turns, `infr run`/`serve` with `INFR_METAL` construct the session-based chat (constrained decode included). Two-turn chat: turn 2 prefills 24 tokens at 19 ms TTFT instead of the full history; decode also gains from the persistent backend (147 → 160 t/s in run mode).

Note for #10 (also on this branch's base): its head gained the rebase + the seam-replay-contract fix + two style commits — see the note there. fmt/clippy/tests green locally on both branches.

digitaloten:
Q8_0 KV follow-ups landed (rebased over the merged #10 first — thanks for resolving the third drift yourself):

- **q8 decode replay tape**: the replay gates accept Q8_0 and pick the q8 dynamic-pos kernels; token streams verified identical replay on/off. Short-context q8 decode 99→149 t/s.
- **Cooperative q8 flash** (`attnflash2_q8kv`): flash2's structure with a dequant-staging stage per 64-position KV block (K pre-transposed [hd][64] in a 24 KB threadgroup tile, V staged into the same tile during the softmax phase). q8 prefill goes from attnvec-class to flash-class: pp8192 1520 t/s (~68% of f16).

Net INFR_KV_Q8 story, measured clean: **0.6B tg128@d16384 = 62.7 t/s (+26% over the f16 cache, +35% over llama.cpp's f16)**; 4B ~−6% (weight-stream-bound — documented as a model-dependent trade); cache footprint −47% everywhere. 46 GPU parity tests, CI matrix green locally.

digitaloten:
One more: **combined_gu implemented on Metal** — the fused [2*nff, ne] gate+up projection your seam capability added. One Linear + GatedActFused per FFN layer instead of two Linears + GatedAct; the op joins the decode-replay whitelist so taped decode keeps recording. Decode +2% (0.6B 156→160 t/s), prefill neutral, output identical (same dots, one tensor). 47 GPU parity tests, CI matrix green locally.

With that, this PR is feature-complete from my side: device MoE (expert-grouped GEMM), persistent Metal KV session, Q8_0 KV cache (write/flash/vec/tape), and combined_gu parity. Happy to mark ready-for-review, split it, or hold for your CI poke — your call.

digitaloten:
Validation sweep on a live gemma checkpoint (gemma-3-1b-it Q4_K_M) — first real gemma through the metal backend, exercising paths previously covered only by synthetic parity shapes: short-prompt CPU-oracle identity, a 1400-token prompt with the sliding window actively clipping (window 512, token-identical vs CPU, correct long-range recall), Q8_0 KV on gemma (hd=256 → the scalar q8 route, identical), the untaped replay path (hd=256 has no tape instantiation — the positions-override contract fix verified live), and the multi-turn Metal session (turn 2 prefilled the 14-token suffix only). **Zero defects found, no code changes.** Noting for completeness: hd=256 has no flash/vec instantiation so gemma rides the split fallbacks — a known instantiation gap, not a correctness issue.

mxaddict:
Merged manually at 6a7e9d0 (main had moved under you again — cmd_serve's Metal arm was rebuilt on the new no-Option seam default with MetalSeamChat behind cfg(macos), and cpu_model keeps both main's multi-slot DenseVulkanSession and your DenseMetalSession). Two review findings fixed in the follow-up commit on main: (1) the kv_q8 gate matched "cpu-reference" but the CPU backend's name() is "cpu" — INFR_KV_Q8 never engaged on the CPU reference, so your q8 CPU-oracle runs actually compared Metal-q8 against CPU-f16 (that they still matched token-for-token is a nice accuracy datapoint, but worth re-running the parity with the gate fixed); (2) `let f16d = kvd;` flipped the roped q16/k16 scratch decls to Q8_0 along with the caches — only the caches quantize now (the Vulkan kv-write peephole fuses on the f16 decl, so the lie was latent but load-bearing-adjacent). Impressive campaign — device MoE, sessions, q8 KV and combined_gu all in; CI green including macOS.

PR #14, opened 2026-07-03
perf(metal): hd=256 flash/vec kernels (gemma 3x) + honest bench sessions

digitaloten:
Ran the new `infr compare --sweep` on Metal (5 models) and it found two craters: **gemma3-1b at 0.22×–0.47×** of llama.cpp across all metrics, and **Q8_0 checkpoints at 0.23×** prefill. Two independent causes, both fixed here.

## 1. hd=256 had no flash/vec instantiations

Every gemma attention (hd=256) rode the split fallbacks. Added:
- `attnflash2_f16kv_hd256` (prefill) — same NSG=4 cooperative structure, 8 O fragments per simdgroup (the `lo[4]` register array was silently sized for hd≤128; now `lo[no]`), 14 KB threadgroup.
- `attnvec_{f16,q8}kv_hd256` + `dyn` variants (decode + replay tape) at **NSG=16**: the per-simdgroup O partials are `NSG·hd·4` bytes, and 32 simdgroups at hd=256 is 32 KB before sq/ssc — over budget. Routing and the replay gates now carry the head-size-dependent threadgroup requirement (512 vs 1024).
- The wide-launch gate opens for hd=256 only when flash2 itself can run (no single-simdgroup flash fallback exists past hd=128).

## 2. bench_metal rebuilt the backend every rep

Per-backend one-time costs — the dequant/repack weight caches, pipeline compiles — were re-paid **inside every measurement**. Invisible on Q4_K_M (native-kernel formats never repack), but a factored-format checkpoint re-repacked hundreds of MB per rep: gemma pp512 spent 687 ms/forward on host encode vs 203 ms of GPU. The bench now runs through one persistent `DenseMetalSession` (weights/pipelines/caches warm — llama-bench keeps one context across reps too), with `SeamKv::reset_tokens()` so each rep still measures a full prefill. Weight caches also key on the underlying MTLBuffer contents pointer rather than the rebindable wrapper address.

## Numbers (M3 Pro, vs llama.cpp Metal, same GGUFs, back-to-back)

| model | metric | before | after | vs llama.cpp |
|---|---|---|---|---|
| gemma3-1b Q4_K_M | pp512 | 673 | **2101** | 0.22× → 0.67× |
| gemma3-1b Q4_K_M | tg128 | 56 | **78.6** | 0.47× → 0.67× |
| gemma3-1b Q4_K_M | tg64@d4096 | 36 | **74.8** | 0.33× → 0.67× |
| qwen3-0.6b Q8_0 | pp512 | 1062 | **3003** | 0.23× → 0.64× |
| qwen3-0.6b Q8_0 | tg128 | 82 | **108** | 0.54× → 0.71× |

Qwen Q4_K_M also picks up ~8% pp512 from the warm-session bench (3622 → 3907).

## Validation

Long-prompt gemma (1400 tokens, window actively clipping) stays SWA-correct vs the CPU oracle — one near-tie greedy flip ("fox" vs "foxes."), the accumulation-order class we've characterized on qwen essays, both outputs verified coherent; replay on/off token-identical. 57 GPU parity tests (3 new hd=256: flash2 windowed, vec windowed at 1e-4, q8 vec at 1e-4). fmt/clippy(-D warnings)/test matrix green locally.

**Stacked on #13** (same exec.rs region — the diff collapses when that merges). Remaining gemma/Q8_0 gap is now the same GEMM-shape story as the dense models (narrow-n / split-K tile variants, the follow-up from #13's notes).

Also noting: `compare --sweep` defaults `--dev Vulkan0`, which llama-bench rejects on a Metal box — ran with `--dev MTL0`. A platform-aware default might be worth a line.

### DIFF
--- a/crates/infr-cli/src/main.rs
+++ b/crates/infr-cli/src/main.rs
@@ -752,18 +752,32 @@ fn cmd_bench_metal(
     {
         let model = infr_llama::CpuModel::load(gguf, tok)?;
         let measure_tg = pg.is_none() && n_gen > 0;
+        // ONE session for warmup + every rep: backend, uploaded weights, compiled pipelines and
+        // the dequant/repack weight caches persist (each rep still measures a full prefill —
+        // `bench_metal` resets the materialized tokens). A fresh backend per rep re-paid those
+        // one-time costs inside the measurement.
+        let ctx = pg.map(|(p, g)| p + g).unwrap_or(if measure_tg {
+            depth.max(1) + n_gen
+        } else {
+            n_prompt
+        }) + 2;
+        let mut sess = model.metal_session(ctx)?;
         // One untimed warmup: page-cache the mmap + build the weight caches + compile pipelines.
-        let _ = model.bench_metal(depth.max(1), if measure_tg || pg.is_some() { 1 } else { 0 });
+        let _ = model.bench_metal(
+            &mut sess,
+            depth.max(1),
+            if measure_tg || pg.is_some() { 1 } else { 0 },
+        );
         let mut samples = Vec::with_capacity(reps);
         for _ in 0..reps {
             let ts = if let Some((p, g)) = pg {
-                let s = model.bench_metal(p, g)?;
+                let s = model.bench_metal(&mut sess, p, g)?;
                 (p + g) as f64 / (s.prompt_secs + s.decode_secs)
             } else if measure_tg {
-                let s = model.bench_metal(depth.max(1), n_gen)?;
+                let s = model.bench_metal(&mut sess, depth.max(1), n_gen)?;
                 n_gen as f64 / s.decode_secs
             } else {
-                let s = model.bench_metal(n_prompt, 0)?;
+                let s = model.bench_metal(&mut sess, n_prompt, 0)?;
                 n_prompt as f64 / s.prompt_secs
             };
             samples.push(ts);
--- a/crates/infr-llama/src/cpu_backend.rs
+++ b/crates/infr-llama/src/cpu_backend.rs
@@ -403,6 +403,13 @@ impl SeamKv {
         common_prefix_len(&self.cached, prompt)
     }
 
+    /// Forget the materialized tokens WITHOUT dropping weights or buffers: the next call
+    /// re-prefills from position 0 into the same session. Bench reps use this so each rep
+    /// measures a full prefill while weights/pipelines/repack caches stay warm.
+    pub(crate) fn reset_tokens(&mut self) {
+        self.cached.clear();
+    }
+
     /// Number of token ids materialized in this slot's KV cache.
     pub(crate) fn cached_len(&self) -> usize {
         self.cached.len()
--- a/crates/infr-llama/src/cpu_backend.rs
+++ b/crates/infr-llama/src/cpu_backend.rs
@@ -406,6 +406,8 @@ impl SeamKv {
     /// Forget the materialized tokens WITHOUT dropping weights or buffers: the next call
     /// re-prefills from position 0 into the same session. Bench reps use this so each rep
     /// measures a full prefill while weights/pipelines/repack caches stay warm.
+    /// (cfg-gated with its only caller, the Metal bench session — dead code on other targets.)
+    #[cfg(target_os = "macos")]
     pub(crate) fn reset_tokens(&mut self) {
         self.cached.clear();
     }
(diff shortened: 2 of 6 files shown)


digitaloten:
Canonical post-fix sweep (idle M3 Pro, `infr compare --sweep --dev MTL0`, r=3):

```
model                   metric     |  infr | llama.cpp | ratio
Qwen3-0.6B-Q4_K_M       pp512      |  3907 |      4295 | 0.91x
Qwen3-0.6B-Q4_K_M       tg128      |   158 |       199 | 0.80x
Qwen3-0.6B-Q4_K_M       tg64@d4096 |   107 |       124 | 0.86x
Qwen3-4B-Q4_K_M         pp512      |   544 |       644 | 0.84x
Qwen3-4B-Q4_K_M         tg128      |    41 |        46 | 0.89x
Qwen3-4B-Q4_K_M         tg64@d4096 |    34 |        37 | 0.90x
qwen3-0.6b-Q8_0         pp512      |  2964 |      4540 | 0.65x
qwen3-0.6b-Q8_0         tg128      |   106 |       148 | 0.71x
qwen3-0.6b-Q8_0         tg64@d4096 |    78 |        99 | 0.79x
qwen3moe-2.4b-Q4_K_M    pp512      |  2314 |      2696 | 0.86x
qwen3moe-2.4b-Q4_K_M    tg128      |   113 |       144 | 0.78x
qwen3moe-2.4b-Q4_K_M    tg64@d4096 |    85 |        95 | 0.89x
gemma3-1b-Q4_K_M        pp512      |  2127 |      3113 | 0.68x
gemma3-1b-Q4_K_M        tg128      |    79 |       118 | 0.67x
gemma3-1b-Q4_K_M        tg64@d4096 |    75 |       115 | 0.65x
```

Worst ratio across the board went 0.22× → 0.65×. The dense Q4_K_M models sit at 0.80–0.91×; the remaining gemma/Q8_0 band (0.65–0.79×) is weight-stream/GEMM-shape work: llama.cpp's Q8_0 GEMV+GEMM run its fastest format while our factored quik8 stream is ~8.5 bpw against their 8.0 packed, and gemma decode is now bound on the same small-model dispatch overheads as qwen 0.6B rather than attention.

mxaddict:
@digitaloten CI is failing 😀

digitaloten:
Second lever landed on this branch — **native Q8_0 kernels** (the other half of the sweep's worst band):

Q8_0 was the one shipped format still on the factored path: ~10.2 bpw streamed + two-level scale indirection vs the raw block's 8.5, and both decode GEMV and prefill GEMM dequant are bound on that stream. Added `DEC16_Q8_0` into the shared RT/CMM/HGEMM shapes plus a `mul_mv_q8_0`-shape GEMV (N_R0=4 rows per simdgroup, llama.cpp's structure) with the fused-residual variant joining the Linear+Add peephole.

Qwen3-0.6B Q8_0, back-to-back vs llama.cpp on the same GGUF:

| metric | before | after | ratio |
|---|---|---|---|
| pp512 | 2964 | **3883** | 0.65× → 0.86× |
| tg128 | 106 | **129** | 0.71× → 0.87× |
| tg64@d4096 | 78 | **90** | 0.79× → 0.91× |

gemma3-1b picks up +4–6% from its Q8_0 tensors (pp512 2209, tg128 83.5). Token streams identical vs the CPU oracle (short + 1400-token SWA prompts), replay on/off identical. 60 GPU parity tests, CI matrix green locally.

With this, every model in the sweep sits at 0.67–0.91× with no outlier band left; the remaining spread is the GEMM-shape / small-model dispatch story.

digitaloten:
CI green (43146cd) — the failure was Linux-only dead code: `SeamKv::reset_tokens`'s single caller is the `cfg(macos)` bench session, so the Linux clippy job saw an unused method the macOS-side checks can't. Gated it with its caller. Ready for review: hd=256 kernels, honest bench sessions, native Q8_0.

digitaloten:
One more on this branch — **concurrent decode replay**:

The sweep's last pattern was short-context decode at 0.78–0.87× across every model while depth decode sits at 0.86–0.91× — depth-independent, so per-token serial cost rather than bandwidth. A dispatch-chain probe (`dispatch_overhead.rs::dispatch_concurrency`) measured the serial encoder at ~1.3 µs/dispatch vs ~0.5 µs concurrent with grouped barriers: `MTLDispatchTypeSerial` implicitly orders every dispatch after the previous one, draining the pipeline between runs of independent work (q/k/v GEMVs, the two KV writes).

The replay tape now encodes with `MTLDispatchTypeConcurrent` and derives exact barrier placement from recorded per-dispatch write masks (RAW/WAW/WAR at buffer granularity). Un-annotated dispatches record writes-everything and replay serialized — never under-fenced. The normal graph walk keeps the serial encoder; only the taped decode hot loop changes.

tg128: 0.6B q4km 158 → 164, q8 129 → 133, gemma 83.5 → 85 (+2–4%); prefill/depth unchanged; token streams identical replay on/off. 60 parity tests, CI matrix green locally (including the Linux dead-code angle this time — the probe test is `#[cfg(target_os = "macos")]`).

PR #15, opened 2026-07-03
perf(metal): native Q5_0 kernels + concurrent decode replay + on-device Copy

digitaloten:
Two levers from the post-#14 board, plus one commit that missed the #14 merge window.

## Native Q5_0 kernels — gemma prefill +18%

gemma Q4_K_M ships its q/k/v/gate/up projections as **Q5_0** (117 of its tensors — ne=1152 isn't a multiple of 256, so K-quant superblocks can't cover those axes), and Q5_0 rode the factored path at ~8.6 bpw streamed vs the raw block's 5.5. Same treatment as #14's Q8_0: `DEC16_Q5_0` into the shared RT/CMM/HGEMM shapes, a mul_mv-shape GEMV (4 rows/simdgroup, 8 blocks in flight, nibble-half-uniform qh extraction per thread), fused-residual peephole variant, w_off slice binds.

gemma3-1b vs llama.cpp Metal, same GGUF: **pp512 2209 → 2606 (0.71× → 0.84×)**, tg64@d4096 74.8 → 82.2 (+10%), tg128 +2% (decode at this size is latency-bound, not stream-bound). Qwen models unchanged.

## Concurrent decode replay (cherry-picked — pushed to #14 right after the merge)

Short-context decode sat at 0.78–0.87× across every model while depth decode was 0.86–0.91× — per-token serial cost, not bandwidth. A probe (`dispatch_overhead.rs::dispatch_concurrency`) measured the serial encoder at ~1.3 µs/dispatch vs ~0.5 µs concurrent with grouped barriers. The replay tape now encodes `MTLDispatchTypeConcurrent` with exact RAW/WAW/WAR barriers derived from per-dispatch write masks recorded at tape time (`TapeEntry.wmask`); un-annotated dispatches record writes-everything and replay serialized — never under-fenced. Normal graph walk unchanged. tg128 +2–4% across models (0.6B 158 → 164); token streams identical replay on/off.

## On-device Op::Copy

The prefill graph extracts the last row's hidden state for the LM head via `Op::Copy`, which was host-side on Metal — a GPU readback + command-buffer break every prefill chunk. Now the rows=1 case of `copy_strided_f32`.

## Validation

64 GPU parity tests (4 new Q5_0: GEMV tail rows, HGEMM, coop-GEMM, add-fusion), long-prompt gemma SWA-correct vs the CPU oracle, replay on/off identical, fmt/clippy(-D warnings)/test matrix green locally including the Linux angle.

### DIFF
--- a/crates/infr-metal/Cargo.toml
+++ b/crates/infr-metal/Cargo.toml
@@ -18,5 +18,10 @@ bytemuck.workspace = true
 metal.workspace = true
 objc.workspace = true
 
+# objc's `sel_impl!` (inside `msg_send!`) expands a `cfg(feature = "cargo-clippy")` check the
+# crate doesn't declare; register it so `-D warnings` builds don't reject `unexpected_cfgs`.
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(feature, values("cargo-clippy"))'] }
+
 [dev-dependencies]
 infr-cpu.workspace = true
--- a/crates/infr-metal/src/exec.rs
+++ b/crates/infr-metal/src/exec.rs
@@ -8,10 +8,12 @@ use infr_core::graph::{Op, TensorKind};
 use infr_core::tensor::{DType, TensorId};
 use infr_core::Result;
 use infr_gguf::dequant::dequant_block;
+use metal::foreign_types::ForeignType;
 use metal::{
     Buffer as MtlBuffer, CommandBuffer, ComputeCommandEncoder, ComputePipelineState,
     MTLResourceOptions, MTLSize,
 };
+use objc::{msg_send, sel, sel_impl};
 use std::ffi::c_void;
 use std::sync::Arc;
 
@@ -118,6 +120,10 @@ pub(crate) struct TapeEntry {
     /// Per-buffer byte offset for `set_buffer` (parallel to `bufs`; all zero except the
     /// fused-QKV weight slices — see the Linear arm's `w_off`).
     offs: Vec<u64>,
+    /// Bitmask over `bufs`: which buffers this dispatch WRITES. Replay runs a CONCURRENT
+    /// encoder and derives the exact barrier placement from these (see `replay_tape`); sites
+    /// that don't annotate record `u32::MAX` — treated as writing everything, i.e. serialized.
+    wmask: u32,
     params: Vec<u8>,
     threads: usize,
     tg: usize,
@@ -353,6 +359,19 @@ impl MetalBackend {
         self.encode_tg(r, pso, bufs, params, threads, 0);
     }
 
+    /// `encode` with a write mask — see `encode_tg_w`.
+    fn encode_w(
+        &self,
+        r: &mut Resident,
+        pso: &ComputePipelineState,
+        bufs: &[&MtlBuffer],
+        wmask: u32,
+        params: &[u8],
+        threads: usize,
+    ) {
+        self.encode_tg_w(r, pso, bufs, wmask, params, threads, 0);
+    }
+
     /// As `encode`, but with an explicit threadgroup width (`tg`; 0 = auto). The simdgroup-GEMV
     /// kernels pass 32 (one simdgroup per threadgroup) so a matvec launches `out_f` threadgroups
     /// instead of a handful of wide ones — far more threadgroups for the GPU's cores to interleave,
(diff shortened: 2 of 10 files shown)


digitaloten:
One more replay lever on this branch — **MoE decode joins the tape** (2322d10):

MoE decode was the last untaped path: `replay_shape` rejected `MoeFfn`, so every token re-walked the graph even though the device MoE path is fully tape-compatible (router GEMV → on-device top-k → expert-batched GEMVs off the device table → reduce; all data dependence resolved on GPU, dispatch args fixed). Whitelisted under the same gate as the arm's device path (the host fallback computes per token and can't tape), with write masks on the seven dispatches so the concurrent replay overlaps the independent gate/up expert GEMVs.

Qwen3-MoE-2.4B: **tg128 114.7 → 122.4 (+7%, 0.78× → 0.85×)**, tg@4096 85 → 87, prefill unchanged. Replay on/off token-identical; 64 parity tests; matrix green locally including the `RUSTFLAGS=-D warnings` angle.

digitaloten:
Final two levers on this branch, plus the plateau sweep:

**Masked-FMA Q5_0 GEMV** (d6fb19c): the chained-GEMV probe put the Q5_0 GEMV at ~39 GB/s vs Q6_K's ~95 — and gemma's gate/up (its two largest per-layer streams) are Q5_0. Rewrote the inner loop q4k-body style (register-side extraction, byte-masked FMAs with exact power-of-2 scale folding, the −16 offset factored through sumy). gemma tg128 86.7 → 93.4.

**k-split GEMV variants** (9d129e2): all four native GEMV bodies gain a compile-time NSG — the `_ks` variants run 4 cooperating simdgroups per row group over strided k-quarters with a threadgroup reduce. Gated on measured evidence both ways: row groups ≤ 4096 AND k deep enough to feed every simdgroup twice (ungated it cost 0.6B −8% on its 4-superblock in_f). 4B tg128 +4%, gemma +2%, MoE +1.4%.

**Canonical sweep (idle M3 Pro, r=3):**

```
model                   pp512   tg128   tg@4096
Qwen3-0.6B-Q4_K_M       0.89x   0.85x   0.90x
Qwen3-4B-Q4_K_M         0.93x   0.89x   0.87x
qwen3-0.6b-Q8_0         0.83x   0.84x   0.90x
qwen3moe-2.4b-Q4_K_M    0.87x   0.85x   0.91x
gemma3-1b-Q4_K_M        0.85x   0.80x   0.77x
```

Floor 0.77× (this arc started at 0.22×). Also ran three further experiments that measured ≤±1.5% and were reverted rather than shipped: an NSG=2 mid-k split, vectorized DEC16 dequant staging for the GEMM path, and (analytically) wider k-split thresholds — decode is now at the weight-stream + dispatch-latency floor, and the prefill band is the GEMM tile-shape story where the cooperative tile already sits at 0.85–0.93×. Diminishing returns from here without new evidence; gemma's depth-decode residual (0.77×) is small-op latency + its 302 MB Q8_0 LM head, both of which llama.cpp also pays — their edge there is kernel-launch density, not bandwidth.

digitaloten:
Post-plateau addendum — wired a **GPU-counter profiler** (`INFR_METAL_PROFILE=3`, 95ec638) and its first reading paid for itself immediately.

**The profiler**: stage-boundary `MTLCounterSampleBuffer` timestamps, one encoder per op inside ONE command buffer — true in-context per-op GPU time. It also exposed that every previous per-op reading had only profiled the RECORDED decode token (the backend tape replays the rest without walking ops — =3 disables the tape), and that metal-rs 0.33's `resolve_counter_range` returns uninitialized memory (worked around with a direct NSData read).

**First catch** (64e74b0): gemma decode spent **20.3% of GPU time in RmsNorm** — 105 single-simdgroup launches/token at ~20 µs each, latency-bound on dim/32 serial loads with nothing else to run at rows==1. `rmsnorm_wide_f32`/`qknormrope_wide_f32` put 8 simdgroups on the row (threadgroup-folded partials), routed for rows ≤ 4.

**Fleet effect — decode is now at parity:**

```
model                   pp512   tg128   tg@4096
Qwen3-0.6B-Q4_K_M       0.89x   0.98x   1.00x
Qwen3-4B-Q4_K_M         0.93x   1.00x   0.96x
qwen3-0.6b-Q8_0         0.84x   0.97x   0.99x
qwen3moe-2.4b-Q4_K_M    0.87x   0.96x   0.99x
gemma3-1b-Q4_K_M        0.84x   0.96x   0.92x
```

tg128 gains: 0.6B +16%, gemma +19%, q8 +10%, 4B +10%, MoE +8%. Token streams identical vs CPU, replay on/off identical, 64 parity tests, matrix green.

Counter-profiled prefill too: 80–90% pure GEMM (gemma's Linear time alone exceeds llama.cpp's whole forward), so the remaining 0.84–0.93× prefill band is the cooperative-tile efficiency grind — no more structural finds. With decode at 0.96–1.00×, calling this branch feature-complete.

digitaloten:
One more evidence-driven arc — **the llama-arch / Q4_0 crater** (72c1bdb):

Pulled a TinyLlama Q4_0 (the classic llama-family ship format) for a coverage check and found it at **0.63–0.66×** — three compounding causes, each new to this PR's coverage:

1. **Q4_0 rode the factored path** (~6.1 bpw vs native 4.5) → native kernels, same recipe as Q8_0/Q5_0.
2. **`Op::Rope` was 18.9% of decode** (counter profiler): one thread per (row, head) serially rotating all pairs — a decode row occupied a single simdgroup. Now one thread per pair.
3. **llama-arch decode never taped**: `replay_shape` rejected `Op::Rope`, so llama-family models got neither record-once replay nor the concurrent encoder. Rope already reads the bound positions buffer, so it's replay-safe by construction — whitelisted.

TinyLlama Q4_0: **tg128 103 → 165 (0.63× → 1.02×), tg@4096 91 → 140 (1.02×)**, pp512 1636 → 2125 (0.85×). Decode now ahead of llama.cpp on this model too. Qwen/gemma fleets unchanged. 68 parity tests, matrix green.

Issue #16, opened 2026-07-03
proposal: engine-level speculative decoding on the seam


digitaloten:
With Metal decode at 0.96–1.02× of llama.cpp (PR #15) and prefill at the measured tile floor, the next decode step-change is algorithmic, not kernel-level — so proposing before building, since this lands in the engine and benefits every backend.

## Sketch

Classic draft-verify on the agnostic seam:

- **Draft**: a small same-tokenizer model (Qwen3-0.6B drafting for 4B/8B; the family shares vocab) generates k candidate tokens per step on the same backend session.
- **Verify**: ONE prefill-shaped forward of the target model over the k candidates — rows=k through the existing batched path. This is where our backends are strongest (the whole prefill machinery: cooperative GEMM, batched attention), and the acceptance loop is pure engine code.
- Accepted prefix commits to the KV cache (WriteKv already batches rows); first rejection resamples from the target logits at that position, standard rejection-sampling correctness — output distribution provably identical to target-only decode, so all the existing token-identity validation machinery keeps working.

## Why the seam makes this cheap

- Sessions already own persistent KV + weights per model; a draft session is just a second (small) session on the same backend instance.
- The verify forward is exactly the batched prefill the backends already run at 0.84–0.93× of llama.cpp — no new kernels anywhere.
- Backend-agnostic by construction: Vulkan gets it the same day Metal does.

## Costs / open questions

- Engine surface: a generate-loop variant that interleaves draft/verify sessions and manages two KV caches; the KV rollback on rejection is a position reset (SeamKv already supports token-list truncation semantics via the suffix-prefill contract).
- Memory: draft model weights + KV resident alongside the target — fine for 0.6B-drafts-4B on 18 GB, tight for bigger pairs.
- Expected gain is workload-dependent (2-3× on greedy/low-temp coding-agent output is the literature number; less on high-entropy text). Worth a prototype gate: measure acceptance length on real transcripts with 0.6B→4B before committing to the full engine surface.

Happy to build it if the design direction works for you — or adjust to whatever engine shape you'd prefer.

digitaloten:
Ran the prototype gate — acceptance-length measurement, zero engine code. `llama-speculative` from brew's llama.cpp measures the same model pairs (acceptance statistics are model-pair properties, engine-independent; temp 0, n=256, three prompt classes: code completion, prose continuation, error-fix explanation).

**Tokens produced per verify round** (accepted + the bonus token — the number a verify forward buys):

[...]

PR #17, opened 2026-07-03
feat(engine): speculative decoding on the Metal seam — draft-verify, token-identical

digitaloten:
Implements #16's design, gated to 8B-class targets by the acceptance measurements there. **Stacked on #15** (uses its session/backend state — diff collapses when that merges).

## Shape

- `INFR_SPEC_DRAFT=<small-gguf>` on `infr run` (Metal, greedy): draft session proposes k tokens (`INFR_SPEC_K`, default 8), ONE batched target forward verifies all of them — the graph builder gains `logits_rows` so verify runs the LM head over every candidate row. Matching prefix commits plus the target's own next token as the bonus.
- **Rollback is free**: the session suffix-prefill contract already handles it — the driver commits only the accepted prefix, and the next round's prefix diff overwrites the speculatively-written KV rows. No new KV machinery.
- **Correctness bar: token-identical to target-only greedy** — holds on every prompt tested (4B and 8B targets).

## Two general backend wins that fell out

- Cooperative-GEMM gate drops m ≥ 16 → m ≥ 2: small multi-row batches (short chat-turn suffix prefills too, not just verify) are weight-stream bound; the tile reads the stream once at GEMM rate where the row-tiled shape re-streams at ~1/3 the bandwidth.
- Split-K cooperative GEMM (`linear_*_cmm_ks` + fixed-order deterministic reduce) for small-m deep-k shapes where out_f/64 threadgroups underfill the GPU.

## Numbers (Qwen3-8B Q4_K_M + 0.6B draft, M3 Pro)

26.7 → **28.9 t/s** at k=12 (+8%), 28.3 at the k=8 default — token-identical output. Honest gap analysis: the acceptance data supports ~2×, and the shortfall is one missing kernel — a **multi-row mul_mv GEMV** for m=2–8 (the verify forward runs ~44 GB/s effective vs the decode GEMV's ~136; counter-profiled, split-K helped only marginally because the tile's serial k-loop latency, not grid width, binds at this size). That kernel is the named follow-up and would also lift short suffix prefills everywhere.

Sessions default to 8k ctx in spec mode (two models + two KV caches share the working set; the full-n_ctx_train pair thrashed a 18 GB machine into second-long forwards — measured before capping).

Draft mode until #15 lands; happy to split the two backend routing commits out if you'd rather take them separately.

### DIFF
--- a/crates/infr-cli/src/main.rs
+++ b/crates/infr-cli/src/main.rs
@@ -894,6 +894,9 @@ fn cmd_bench_qwen35(
     std::env::set_var("INFR_Q35_IGNORE_EOS", "1"); // fixed tg count, no early stop
     let mut m: Box<dyn ChatModel> = if cpu {
         Box::new(infr_llama::model::Qwen35Chat::new_cpu(gguf.to_path_buf()))
+    } else if std::env::var("INFR_METAL").is_ok() {
+        // Metal seam (there is no Vulkan loader on macOS — `new()` would dlopen-fail).
+        Box::new(infr_llama::model::Qwen35Chat::new_metal(gguf.to_path_buf()))
     } else {
         Box::new(infr_llama::model::Qwen35Chat::new(gguf.to_path_buf()))
     };
--- a/crates/infr-cli/src/main.rs
+++ b/crates/infr-cli/src/main.rs
@@ -438,12 +438,13 @@ fn cmd_run(model: &str, message: Option<&str>) -> anyhow::Result<()> {
                     let target = infr_llama::CpuModel::load(&gguf, tok.as_deref())?;
                     let draft =
                         infr_llama::CpuModel::load(std::path::Path::new(&draft_path), None)?;
-                    // Verify cost is nearly flat in k (one batched forward), so a larger k
-                    // amortizes it; 8 balances the draft cost against acceptance decay.
+                    // Upper bound on the draft length; the driver adapts the actual k per
+                    // round to recent acceptance (verify cost scales with rows on this
+                    // hardware, so over-drafting low-acceptance text costs real time).
                     let k = std::env::var("INFR_SPEC_K")
                         .ok()
                         .and_then(|v| v.parse().ok())
-                        .unwrap_or(8);
+                        .unwrap_or(6);
                     let (tb, db) = (
                         std::fs::metadata(&gguf).map(|m| m.len()).unwrap_or(0),
                         std::fs::metadata(&draft_path).map(|m| m.len()).unwrap_or(0),
--- a/crates/infr-cli/src/main.rs
+++ b/crates/infr-cli/src/main.rs
@@ -431,9 +431,39 @@ fn cmd_run(model: &str, message: Option<&str>) -> anyhow::Result<()> {
             );
             #[cfg(target_os = "macos")]
             {
-                Box::new(infr_llama::model::MetalSeamChat::new(
-                    infr_llama::CpuModel::load(&gguf, tok.as_deref())?,
-                ))
+                // INFR_SPEC_DRAFT=<gguf>: speculative decoding — a small same-tokenizer draft
+                // proposes k tokens (INFR_SPEC_K, default 4), one batched target forward
+                // verifies. Greedy-only; pays off for ~8B-class targets (issue #16).
+                if let Ok(draft_path) = std::env::var("INFR_SPEC_DRAFT") {
+                    let target = infr_llama::CpuModel::load(&gguf, tok.as_deref())?;
+                    let draft =
+                        infr_llama::CpuModel::load(std::path::Path::new(&draft_path), None)?;
+                    // Verify cost is nearly flat in k (one batched forward), so a larger k
+                    // amortizes it; 8 balances the draft cost against acceptance decay.
+                    let k = std::env::var("INFR_SPEC_K")
+                        .ok()
+                        .and_then(|v| v.parse().ok())
+                        .unwrap_or(8);
+                    let (tb, db) = (
+                        std::fs::metadata(&gguf).map(|m| m.len()).unwrap_or(0),
+                        std::fs::metadata(&draft_path).map(|m| m.len()).unwrap_or(0),
+                    );
+                    if db * 4 > tb {
+                        eprintln!(
+                            "[spec] warning: draft is more than 1/4 the target's size — \
+                             speculation only pays when the target is much larger (see #16)"
+                        );
+                    }
+                    std::env::set_var("INFR_TEMP", "0");
+                    eprintln!(
+                        "[metal spec — target + {k}-token draft verify, greedy (INFR_TEMP=0)]"
+                    );
+                    Box::new(infr_llama::model::SpecMetalChat::new(target, draft, k))
+                } else {
+                    Box::new(infr_llama::model::MetalSeamChat::new(
+                        infr_llama::CpuModel::load(&gguf, tok.as_deref())?,
+                    ))
+                }
             }
             #[cfg(not(target_os = "macos"))]
             {
--- a/crates/infr-llama/src/cpu_backend.rs
+++ b/crates/infr-llama/src/cpu_backend.rs
@@ -115,6 +115,7 @@ pub(crate) fn generate_dense_cpu(
         &mut None,
         prompt.len() + max_new + 1,
         None,
+        None,
     )
 }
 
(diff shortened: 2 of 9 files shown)


digitaloten:
Follow-up landed (c197cb3) — **multi-row GEMV verify + adaptive draft length**:

The verify bottleneck decomposed into three measured kernel shapes:
1. Register-hoisted multi-row body (weights in registers, token loop inside): **ALU-bound**, ~24 ms per extra row — falsified.
2. **Token-parallel `mrv`** (one simdgroup per (row-pair, token), token as the fastest grid index so same-row simdgroups schedule together and their weight-block reads coalesce in cache): verify m=4 **122 → 80 ms**. Routed for m=2–8 K-quant shapes.
3. Split-K cooperative GEMM keeps m > 8.

Plus an acceptance-EMA **adaptive k** (bounds [1, cap], cap default 6): verify cost scales with rows on this hardware, and fixed-k over-drafting measured net-negative on prose (19.8 vs 26.5 baseline); adaptive holds prose at break-even while keeping the code win.

**Qwen3-8B + 0.6B draft: 26.7 → 30.0–30.9 t/s on code-shaped output (+12–16%), prose break-even, everything token-identical to target-only greedy.** 70 parity tests, matrix green.

Honest ceiling note: the acceptance data supports ~2× if verify were flat-cost in k; the residual is the ~20 ms/row slope of the token-parallel GEMV (simdgroups drift out of lockstep over the block walk — a threadgroup-cooperative scalar stage could pin them but pays barrier latency per block; measured the class, parked).

digitaloten:
Two more coverage arcs on this branch — both found real bugs, one of them silent corruption:

## Qwen3.5-4B (1caff08) — the first qwen35 checkpoint that fits a dev box

The device DeltaNet/Conv1dSilu port from #12 **finally validates e2e**: token-identical vs the CPU oracle on short and 1400-token prompts. Sweep found 0.52–0.76×; three fixes: `cmd_bench_qwen35` never routed to Metal (Vulkan dlopen — the sweep column was ERR), the qwen35 seam declared **F32 KV caches** (every GQA layer fell to the scalar f32 split, 33 ms/call at depth — now F16 like the dense seam), and qwen35 decode never taped (DeltaNet/Conv1dSilu/QkNorm/Scale/Copy/CopyStrided whitelisted with device-gate mirroring + write masks). **Result: tg128 27→34, tg64@d4096 18→33 — 0.52× → 0.95×.**

## IQ4_XS/IQ4_NL (d9aa0b1) — silent corruption on the most common mac format

bartowski Qwen3-4B IQ4_XS produced **garbage** on Metal: codebook quants fell to the dense branch's cached-F32 dequant — a 4B model balloons to 16 GB against a 12.9 GB working set and the allocation pressure corrupts silently (a 0.6B, whose cache fits, was correct — the discriminating experiment). Native `DEC16_IQ4XS`/`DEC16_IQ4NL` (the kvalues codebook as a constant table, bit-exact vs `dequant_codebook`) across all four Linear shapes; plus first-ever parity coverage for Q5_K-factored and IQ4. **Result: garbage @1.9 t/s → token-identical @ 0.89–0.90× of llama.cpp.**

74 GPU parity tests, matrix green. The coverage-model lesson keeps paying: five of six new models/formats found a distinct structural bug.

mxaddict (changes requested):
Reviewed the spec-decode delta (`git diff pr15...pr17`). The **driver itself is correct** — I traced the accept/commit loop (`base + j` predicts `cand[j]`, bonus at `base + cand.len()`), it's off-by-one-free, and the KV rollback is sound: `cached` is truncated to `common_prefix_len` at the first divergence and the suffix-prefill overwrites the rejected (speculatively-written) KV rows before any query attends them. The IQ4_XS/IQ4_NL kernels, qwen35 f16 KV, and e2b prologue also check out. Nice work.

Two things to close before merge, both about the numeric paths that back the headline:

**1. No parity coverage for the small-m coop-GEMM / split-K verify path.**
The gate widening `m >= 16 -> m >= 2` (exec.rs) routes small batches through the new `cmm`/`cmm_ks` (split-K) kernels + `cmm_ks_reduce`. This is not opt-in — it also fires on normal short chat-turn **suffix prefills**, not just spec verify. But every coop-GEMM parity test in `tests/parity.rs` uses `m = 40` (so `ks_split` collapses to 1), meaning `cmm_ks` and the m=2..15 regime are never exercised in CI. The k-partition arithmetic looks OOB-free on read, but a new f16 MMA reduction with no oracle on the hot path needs a test. Please add a parity case at a real verify shape — `m` in 2..7 and `in_f >= 2048` (deep-k, so `ks_split > 1`) — asserting CPU-vs-Metal within tolerance.

**2. \"Token-identical to target-only greedy\" is empirical, not guaranteed — and untested end-to-end.**
Verify runs the batched f16 GEMM/cmm path; target-only greedy decode uses the exact-f32 GEMV (the \"m==1 keeps the exact-f32 GEMV\" contract). The accept loop is bit-identical to greedy-over-the-*verify-forward*, but that can diverge from greedy-over-the-*f32-decode* on a near-tie logit. Please either add an end-to-end test asserting `spec_output == target_only_greedy_output` on a small model, or soften the claim to \"token-identical to the target's batched-verify greedy.\"

**Rebase note:** #15 and #19 are now on `main`. Rebasing #17 onto main will drop its #15 commits (already applied). The `Op::WriteKv` exec.rs conflict is already resolved in main (per-format thread counts + `encode_w` mask); your spec-routing edits to `Op::Linear`/`Op::Attention` may still touch main's new KV-quant prepass arms — worth a careful pass there.

mxaddict:
Heads up: the changes requested here also **block #18**. `metal-multislot` fully contains this branch (`git merge-base --is-ancestor pr17 pr18` = true — #18 is #17 plus the single SlotPool commit), so the multi-slot PR can't merge until the two items above (small-m `cmm_ks`/coop-GEMM parity coverage + the spec-vs-greedy equivalence test / claim softening) are resolved here. Landing those unblocks both PRs at once.

digitaloten:
Both items addressed, and the branch is rebased onto main (9f303df on top).

**1. Split-K / small-m coop-GEMM parity — 4 new tests at real verify shapes** (`linear_*_splitk_verify_shape_matches_dequant_reference`):

| case | m | in_f | out_f | ks_split |
|---|---|---|---|---|
| Q8_0 | 3 | 2048 | 256 | 8 |
| Q5_0 | 5 | 2048 | 256 | 8 |
| Q6K | 9 | 2048 | 256 | 8 |
| Q4K | 12 | 4096 | 512 | 8 |

Shape notes: K-quants at m∈2..=8 route to the multi-row GEMV (covered by the existing multirow tests), so the K-quant split-K cases sit at m∈9..15; the non-K formats cover the m∈2..8 regime you asked for. I verified the ROUTE, not just the result — instrumented the `ks_split > 1` branch and confirmed all four dispatch `cmm_ks` with split=8 (then removed the instrumentation).

One honest finding from writing them: at k=2048 the Q6K case measured 1.4e-3 relative error against the f16-operand reference — the MMA rounds per-product at ~2^-11 and deep k accumulates it, so these four tests use 2.5e-3 (the shallow k=256 tests keep 1e-3). The comment block in parity.rs records the reasoning.

**2. Spec ≡ target-only greedy — now an end-to-end test**, not an empirical claim: `metal_spec_decode_matches_target_only_greedy` asserts the full 64-token spec stream equals target-only greedy on Qwen3-0.6B (self-spec pair — the driver never knows draft and target are the same file, so the accept/commit machinery is exercised identically). It passes. You're right that the guarantee is precisely "greedy over the VERIFY forward," which can split from f32-GEMV decode on a near-tie — the driver doc now states exactly that contract and points at the test that pins the end-to-end equality.

**Rebase**: clean onto current main (#15/#19 + the KV-quant campaign). The spec-routing edits and your Attention prepass compose without conflict — the m≥2 cmm gate only touches `Op::Linear`, and verify's attention rows run through the same prepass path as any prefill. All 88 GPU parity tests green (including the new KV-quant attention cases), full matrix green.

#18 is also rebased: it's now exactly this branch + the single SlotPool commit, per your gating note.

mxaddict (approved):
Both requested items addressed and independently verified: the 4 split-K tests genuinely engage cmm_ks (recomputed ks_split=8 from the gate for all four shapes), and metal_spec_decode_matches_target_only_greedy pins the f16-verify/f32-decode equivalence end-to-end. Rebase is clean, CI green. Approving. (One optional follow-up: the self-spec e2e never takes the rejection branch since acceptance is always full — I'll add a hermetic unit test of the accept/reject logic as a follow-up on main.)

PR #20, opened 2026-07-04
test(metal): coupled quant-KV attention parity at depth

digitaloten:
Companion to the KV-quant validation report (issue to follow, linked below once filed). The coupled quant/quant prepass cases ran only at kv_len=6; this adds one test sweeping all five block formats coupled at kv_len=2048 — the rows==1 decode shape at pos=2047 and an 8-row prefill shape — so the scratch block arithmetic and deep-position reads are pinned.

All ten cases pass within the existing 6e-3 f16-scratch tolerance on an M3 Pro. That result is itself the load-bearing fact in the validation report: it discriminates "kernel indexing bug at depth" (refuted) from "precision compounding" (confirmed) for the coupled-iq4_nl recall gap described there.

fmt / clippy -D warnings / workspace tests / 79 GPU parity green locally.


digitaloten:
Validation report with the full context: #21

Issue #21, opened 2026-07-04
Metal KV-quant prepass: hardware validation report (M3 Pro) — correct at depth, decode craters past ~2k context


digitaloten:
5e7ec86 shipped blind ("the macos-15 CI job is the verifier"), so here's the report from real hardware — M3 Pro 18-core, Qwen3-0.6B Q8_0, greedy, `INFR_KV_TYPE_K/V` per format.

## Correctness: clean. One behavior finding, no kernel bugs.

**e2e sweep, Metal vs the CPU oracle at the same KV format** (short prompt, 32 tokens): f16 / bf16 / f32 / mixed K=f16+V=q4_0 are token-IDENTICAL. The block quants differ token-wise but match the CPU's quality class (coherent where CPU is coherent) — the documented accumulation-order near-tie class. turbo2/3/4 produce garbage on BOTH backends at 0.6B — that's 2–4-bit KV destroying a tiny model, not a kernel defect (worth a README/docs note that turbo is for models with margin to spend).

**The one real finding — coupled 4-bit KV loses long-context recall that the CPU oracle keeps.** Planted-secret probes at ~800 and ~3400 tokens, 3 prompts each: coupled iq4_nl on Metal recalled **0/6** while CPU-coupled-iq4_nl recalled 4/6. Discriminators run:

- Per-side mixes (K=iq4_nl+V=f16, K=f16+V=iq4_nl) at 3400 tokens: **both recall perfectly** on Metal. Only both-sides-quantized fails.
- Coupled parity at depth (kv_len=2048, decode + prefill shapes, all five block formats): **passes within the existing 6e-3 tolerance** — PR #20. So the scratch indexing and quantize/dequant kernels are clean at depth.
- bf16 through the SAME prepass + f16 attention path recalls perfectly, so the f16 attention alone isn't the killer.

Conclusion: 4-bit loss on BOTH sides × f16 attention accumulation eats the recall margin (the CPU oracle attends in f32 over the same quantized values). Not a bug — but it means the llama.cpp guidance the code comment already cites (high-precision K + quantized V) is a *quality requirement* on Metal, not just a suggestion. Might be worth clamping or warning on coupled ≤4-bit K on metal the way mixed-native already clamps.

## Performance: prepass formats crater at depth

tg on 0.6B Q8_0 (r=3 shallow, r=2 deep, idle machine, persistent bench session):

| KV format | tg128 | tg64 @ d4096 |
|---|---|---|
[...]

PR #23, opened 2026-07-04
perf(metal): native-read q4_0/iq4_nl KV decode — 3.2x at depth, replay-taped

digitaloten:
The follow-up offered in #21. The prepass re-expands the entire kv prefix into fresh scratch every Attention call (O(n²) dequant over a generation) and forces static per-token decode. The q8 native-read recipe extends mechanically to the 4-bit blocks: in-kernel block decode, f32 accumulate, replay-taped decode.

## Numbers (Qwen3-0.6B Q8_0 weights, M3 Pro, coupled K=V)

| KV format | tg128 | tg64 @ d4096 |
|---|---|---|
| f16 | 144.9 | 96.1 |
| q8_0 | 138.7 | 107.2 |
| **q4_0** | 120 → **139.7** | 31.7 → **103.9 (3.3×)** |
| **iq4_nl** | 106 → **130.7** | 31.1 → **97.8 (3.1×)** |

q4_0 now beats f16 at depth on half the cache bandwidth. Mixed pairs (K=f16 + quant V etc.) and wide prefill keep the prepass — the vector kernel's per-(row, head) shape is serial over kv and loses to dequant-once + flash there.

## Shape

- The q8 vector-flash body generalizes over a KV accessor struct (`KVQ8`/`KVQ40`/`KVIQ4NL`, exact ports of the CPU dequant formulas); q8's kernels re-instantiate through it with their host names unchanged. New `attnvec_[dyn_]*` instantiations at hd 64/128/256 and `writekv_dyn_q4_0`/`writekv_dyn_iq4_nl` (byte-identical quantization to the static kernels).
- `replay_shape` admits the **coupled** formats only — a mixed pair has no native dyn kernel and keeps static+prepass. Dyn kernel naming is centralized (`dyn_attnvec_kern`) so the recording route and the pipeline-cap check can't disagree.
- One bug caught by the replay-identity bar during development: with a kv_len floor on the static native route, the shallow-kv prefix went through the f16-scratch prepass while the tape's dyn kernel read exactly — replay-on/off decode diverged on early tokens. The native route now has no kv floor; both formats are token-identical replay on/off.

## On the #21 recall finding

Partial honesty update: f32 accumulation over exact decodes (this PR) improves coupled-iq4_nl long-context recall but does **not** fully reach the CPU oracle's rate — the residual is f16 activation rounding *upstream* of the quantizer, inherent to the f16 compute pipeline, not addressable in the attention read.

## Verified

89 GPU parity tests (the deep-coupled kv_len=2048 cases from #20 now exercise the native kernels on both decode and 8-row shapes; routing confirmed by instrumentation), replay on/off token-identical for both formats, f16/q8 paths unchanged, fmt/clippy `-D warnings`/workspace tests green.

### DIFF
--- a/crates/infr-metal/src/exec.rs
+++ b/crates/infr-metal/src/exec.rs
@@ -240,7 +240,10 @@ fn replay_shape(g: &infr_core::graph::Graph, bindings: &Bindings) -> bool {
                 }
             }
             Op::WriteKv { cache, .. } => {
-                if !matches!(g.desc(*cache).dtype, DType::F16 | DType::Q8_0) {
+                if !matches!(
+                    g.desc(*cache).dtype,
+                    DType::F16 | DType::Q8_0 | DType::Q4_0 | DType::Iq4Nl
+                ) {
                     return false;
                 }
             }
@@ -248,13 +251,17 @@ fn replay_shape(g: &infr_core::graph::Graph, bindings: &Bindings) -> bool {
                 rows,
                 head_dim,
                 k_cache,
+                v_cache,
                 ..
             } => {
                 has_attn = true;
-                if *rows != 1
-                    || !matches!(*head_dim, 64 | 128 | 256)
-                    || !matches!(g.desc(*k_cache).dtype, DType::F16 | DType::Q8_0)
-                {
+                let (kdt, vdt) = (g.desc(*k_cache).dtype, g.desc(*v_cache).dtype);
+                // f16/q8 are coupled by the runner's clamp; q4_0/iq4_nl compose per-side in
+                // general but only the COUPLED shape has a native-read dyn kernel — a mixed
+                // pair keeps the static prepass path.
+                let native = matches!(kdt, DType::F16 | DType::Q8_0)
+                    || (kdt == vdt && matches!(kdt, DType::Q4_0 | DType::Iq4Nl));
+                if *rows != 1 || !matches!(*head_dim, 64 | 128 | 256) || !native {
                     return false;
                 }
             }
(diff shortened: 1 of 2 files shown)


mxaddict:
Reviewed — and this PR's content is already on main: commit e73bfa1 is patch-id-identical to f539814 (`git cherry` marks it equivalent; it landed in the direct push batch). Closing rather than merging to avoid an empty merge.

Review of the live code: the KV accessor structs are exact CPU-formula ports (q4_0 `d*(nib-8)`, iq4_nl codebook; nibble-half indexing and 2-byte alignment check out — 18-byte blocks keep `d` aligned), the dyn quantizers are byte-identical to the static kernels, and centralizing `dyn_attnvec_kern` so the recording route and cap-check share one name table is exactly right. Nice catch on the kv_len-floor replay divergence.

**One finding, fixing on main directly:** `replay_shape`'s Attention gate admits a **mixed K=f16 + V=q4_0/iq4_nl** pair as native — `matches!(kdt, F16 | Q8_0)` only inspects K, and both cache dtypes pass the WriteKv whitelist. That pair is reachable (the runner clamp only forces coupling for q8/f32; prepass formats compose per-side — and high-precision-K + quant-V is the recommended footprint shape). The tape then records `attnvec_dyn_f16kv_*`, which reads the q4_0 V cache as f16 → corrupt decode under replay, while the static walk would have prepassed V correctly. Fix: require `kdt == vdt` for every native format (q8/f16 are coupled by the clamp anyway, so no working case loses its tape). Worth a replay-on/off identity run with `INFR_KV_TYPE_K=f16 INFR_KV_TYPE_V=q4_0` on hardware next time you're on the M3.

digitaloten:
Hardware verification of 99b27e1 as requested — replay-on/off decode identity on the M3 Pro (0.6B Q8_0 weights, greedy, 32 tokens):

| pair | replay on vs off |
|---|---|
| K=f16, V=q4_0 | IDENTICAL |
| K=f16, V=iq4_nl | IDENTICAL |
| K=q4_0, V=f16 | IDENTICAL |

All three mixed pairs now correctly take the static prepass path (no tape) and produce identical streams either way. Good catch on the K-only inspection — the coupled requirement is the right shape for the gate.

PR #26, opened 2026-07-04
perf(metal): simdgroup-per-column DeltaNet — qwen35 prefill 0.73x -> 0.80x

digitaloten:
Stacked on #24 (edits `shaders/deltanet.metal`; one commit on top — review 7b96514).

## Evidence chain

Counter profile on qwen35 pp512: Linear 71% (the known GEMM floor), **DeltaNet 23%** (302 ms per chunk set) — the only structural residual. The old kernel ran one threadgroup per value head: 32 threadgroups on Qwen3.5-4B, an order of magnitude under-occupied, and it re-read the state column from device twice per token.

## New shape

The llama.cpp `kernel_gated_delta_net` parallelization: one **simdgroup per (value-dim, head)** — 32 lanes split the k-dim, each lane holding KPL = kd/32 state entries in registers across the whole chunk. The token recurrence needs only `simd_sum`s: no threadgroup memory, no barriers. Four simdgroups share a threadgroup for occupancy (grid = nv·vd/4 threadgroups vs the old nv).

**One instructive failure worth the commit message:** the first version used KPL as a runtime loop bound and measured **507 ms — 1.7× slower than the old kernel**. Runtime-indexed arrays can't be register-promoted, so the "register-resident" state silently spilled to thread-private memory. With KPL as a compile-time template parameter (`deltanet_f32_k1/2/4/8`, host picks by kd/32) and unrolled loops: **185 ms, 1.6× faster than the old shape**.

## Numbers (Qwen3.5-4B, M3 Pro)

- pp512: 428 → **471 t/s (0.73× → 0.80× of llama.cpp)**. The residual is the Linear band at its known floor (78% of GPU time — qwen35 simply runs more projections per token than dense).
- tg128 35.9 (+2%), tg64@d4096 34.2 (unchanged class — decode was already replay-taped).

## Bars

Metal-vs-CPU-oracle token identity holds; replay-on/off identity holds; deltanet parity at 1e-4 (simd_sum tree reassociation vs the CPU's serial sum); full matrix + 89 GPU parity green.


digitaloten:
Rebased onto #27 (this PR's kernel edit was INERT against current main — the shaders/*.metal files are orphaned there; see #27 for the regression). With the shader source live again: qwen35 pp512 462.5 t/s re-verified on the restacked branch (0.79-0.80x), token identity vs the CPU oracle holds. Review order: #27 first, then this collapses to the one DeltaNet commit.

PR #32, opened 2026-07-05
fix(vulkan): enumerate portability drivers + probe features instead of assuming them

digitaloten:
Found while attempting a MoltenVK test rig on the M3 (goal: local Vulkan-side validation next to Metal). Two init assumptions only a native driver satisfies:

1. **Portability drivers are hidden by the loader** unless the instance enables `VK_KHR_portability_enumeration` + `ENUMERATE_PORTABILITY_KHR`. Without it, a Mac with MoltenVK installed dies at `create_instance: Unable to find a Vulkan driver`. Now probed and opted into (no-op on native-driver platforms), plus `VK_KHR_portability_subset` on the device when advertised — a valid-usage *requirement* for layered devices.
2. **`vulkanMemoryModel` and `subgroupSizeControl` were enabled unconditionally** — enabling an unsupported feature fails `create_device` outright. Both now come from the `features2` probe, like `shaderFloat16` already did.

Honest scope note: with this, MoltenVK enumerates and instance+device creation succeed on the M3, but the backend then hits the real wall — some SPIR-V pipelines fail `ERROR_INITIALIZATION_FAILED` under MoltenVK's MSL translation. So this is the necessary first step of MoltenVK support, not the whole story; I timeboxed rather than chasing shader-by-shader compatibility (the native Metal backend exists). Filed as-is because the fixes are correct independent of MoltenVK: any layered/portability driver and any device lacking those two features hits these paths.

On native-driver platforms nothing changes — same extensions, same features, now conditioned on what the device actually reports. fmt/clippy/workspace tests green.

### DIFF
--- a/crates/infr-vulkan/src/lib.rs
+++ b/crates/infr-vulkan/src/lib.rs
@@ -312,9 +312,28 @@ impl VulkanBackend {
             .engine_version(vk::make_api_version(0, 0, 1, 0))
             .api_version(vk::API_VERSION_1_3);
 
+        // Portability drivers (MoltenVK on macOS) are HIDDEN by the loader unless the instance
+        // opts in with VK_KHR_portability_enumeration + the matching create flag — without it,
+        // create_instance on a Mac reports "unable to find a Vulkan driver" even with MoltenVK
+        // installed. Probe and opt in when available; a no-op on platforms with native drivers.
+        let inst_exts =
+            unsafe { entry.enumerate_instance_extension_properties(None) }.unwrap_or_default();
+        let has_portability = inst_exts.iter().any(|e| {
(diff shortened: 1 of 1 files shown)


mxaddict:
Merged. Post-merge validation on the 7900 XTX / RADV: full GPU test suite 39/39 green, coopmat GEMM tier still engaged (pp512 15.3k t/s on the 0.8B) — the probes resolve to the previously-hardcoded values on a native driver, as advertised. Also landed the follow-up from review: the cooperative-matrix FEATURE bit is now probed too (same failure class — ext-present-but-feature-false would fail create_device), which should be one less landmine for the MoltenVK rig when you get past the MSL translation wall.

Issue #33, opened 2026-07-05
MTP (multi-token prediction) speculative decoding — qwen35 single-head first


mxaddict:
llama.cpp master now ships MTP as a speculative type (`--spec-type draft-mtp`; `--mtp` downloads the head), and two of the three MTP-enabled arch families are ones infr supports: **qwen35** (single trained MTP head) and **gemma4** (memory-shared mode — the head shares the target KV and runs in one graph). The MTP head is published as a **separate sibling GGUF** (`mtp-*.gguf`) in the HF repos, not baked into the main file.

Reference: `common/speculative.cpp` `common_speculative_impl_draft_mtp` in llama.cpp (three driver modes: `is_mem_shared` = gemma4, `chain_heads` = step35, single-head = qwen35/qwen35moe), `llama_model_n_layer_nextn`, the `nextn.*` tensor family (`eh_proj`, `embed_tokens`, `enorm`/`hnorm`, `shared_head_{head,norm}`, `pre/post_projection`), and the `llama_set_embeddings_nextn` staging API that feeds the head the target's hidden states.

## Scope — phase 1: qwen35 single-head

The verify side already exists in infr (`spec_accept`, batched verify forwards, multi-slot spec serve). The delta is the draft side:

1. **Head loading**: resolve + pull the sibling `mtp-*.gguf` (infr-hub already handles sibling files for mmproj-style cases — check), parse the `nextn.*` tensors.
2. **Hidden-state tap**: the MTP head consumes the target's final hidden state h_p paired with the next token embedding x_{p+1} — the seam needs an opt-in output of the pre-lm_head hidden row(s) from a forward (we already do multi-row logits for spec verify; this is the same plumbing one op earlier).
3. **Draft loop**: token-by-token through the single head (embed x → concat/project with h via `eh_proj` → one transformer layer → shared head → logits), k drafts, then the existing verify.
4. **Cross-batch carryover**: the last verified row's h pairs with the NEXT call's first token (llama.cpp stashes `pending_h` per seq — same trick).

Bar: spec ≡ target-only greedy (the invariant the existing spec tests pin), measured acceptance rate + tok/s vs the two-model draft on the same target.

## Later phases
- **gemma4 memory-shared mode** — most interesting perf-wise (no separate KV, one graph).
- qwen35moe when the arch lands.
- `draft-dflash` exists in llama.cpp too now — related but separate (issue #16 already has the comparison note).

Queued after DG perf Phase D (Metal Canvas attention).

[...]

PR #36, opened 2026-07-05
feat(mtp): Metal + CPU MTP drivers — correct on both, Metal head-numerics gap flagged

digitaloten:
Extends the MTP arc (issue #33) off the Vulkan-only path. The head session + the whole draft-verify-catchup loop were already backend-generic (`MtpHeadSession` holds a `&dyn Backend`; `run_verify` is just `generate_dense_backend`), so only three spots hardcoded Vulkan. Refactored those into a backend-generic core (`generate_mtp_spec_core`) with thin `_vulkan`/`_metal`/`_cpu` wrappers, added `MtpHeadSession::new_metal`, and wired `INFR_MTP` into `MetalSeamChat` + `CpuDenseChat`.

## Validation (Qwen3.5-4B-MTP UD-Q4_K_XL, M3 Pro)

**Correctness — solid on both new backends.** MTP output is token-identical to target-only greedy, every prompt tried (the same invariant the Vulkan path is pinned on). Verified on Metal and CPU.

**Perf — a Metal-specific gap, and the reason this lands "correct" not "fast":**

| backend | precision | alpha (accept rate) |
|---|---|---|
| CPU | exact f32 | **0.96** |
| Vulkan (reference) | f16 | ~0.95 |
| **Metal** | f16 | **0.26** |

CPU's 0.96 matches the Vulkan reference, so the **port logic is correct**. Metal's 0.26 is far lower — and since Vulkan f16 also hits ~0.95, it's **Metal-specific, not generic f16 rounding**. The trunk verify on Metal is already proven correct (spec-decode is token-identical there), so the divergence is in the **head forward's Metal execution**. At alpha 0.26 every reject triggers a full reprefill (qwen35 no-rewind), so MTP is currently **net-negative on Metal** — it needs the head numerics pinned before it's a win.

The `generate_mtp_spec_cpu_timed` driver is the exact-f32 acceptance oracle that made that diagnosis possible; it stays as the reference for the head-numerics follow-up (and is useful on its own).

## How to read this PR

Two things bundled: (1) the backend-generic refactor + CPU/Metal wiring (mergeable — correct, non-MTP paths untouched, `INFR_MTP` unset is a no-op), and (2) a precisely-localized Metal head-numerics bug handed back with an isolating experiment. If you'd rather I keep digging the head's per-op Metal numerics before this lands, say the word — I have the model cached and the CPU oracle to A/B against. Happy to split the refactor from the Metal path if you want the CPU oracle in now and the Metal path held.

Non-MTP models/paths untouched. Full matrix + 92 GPU parity green.

(Separate, noticed while validating: `DiffusionSess` trips clippy's `large_enum_variant` on macOS — the `#[cfg(macos)]` Metal variant makes the size spread — but the Linux clippy CI job never compiles that variant, so it's invisible there. Filing that separately.)

### DIFF
--- a/crates/infr-llama/src/model.rs
+++ b/crates/infr-llama/src/model.rs
@@ -452,6 +452,8 @@ impl ChatModel for DenseSeamChat {
 pub struct MetalSeamChat {
     model: CpuModel,
     session: Option<crate::cpu_model::DenseMetalSession>,
+    mtp_head: Option<crate::mtp::MtpHeadWeights>,
+    mtp_checked: bool,
 }
 
 #[cfg(target_os = "macos")]
@@ -460,7 +462,32 @@ impl MetalSeamChat {
         Self {
             model,
             session: None,
+            mtp_head: None,
+            mtp_checked: false,
         }
+    }
+
+    /// MTP mode is opt-in (`INFR_MTP=1`) and only for a qwen35 GGUF that ships an MTP head — the
+    /// Metal twin of [`DenseSeamChat::wants_mtp`]. Memoized so a non-MTP GGUF doesn't re-parse
+    /// its `Config` every turn.
+    fn wants_mtp(&mut self) -> Result<bool> {
+        if self.mtp_head.is_some() {
+            return Ok(true);
+        }
+        if self.mtp_checked {
+            return Ok(false);
+        }
+        self.mtp_checked = true;
+        if std::env::var("INFR_MTP").ok().as_deref() != Some("1")
+            || self.model.config().n_layer_nextn == 0
+        {
+            return Ok(false);
+        }
+        self.mtp_head = Some(crate::mtp::load_mtp_head(
+            self.model.gguf(),
+            self.model.config(),
+        )?);
+        Ok(true)
     }
 
     fn ensure_session(&mut self) -> Result<()> {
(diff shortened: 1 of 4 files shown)


digitaloten:
**Root cause found and fixed — it wasn't head numerics, it was the decode-replay tape (#38).**

I dug the alpha-0.26 gap to the bottom. Isolation chain: every head op is parity-correct on Metal (hd=256 short-kv attention and hd=256 partial rope both pass new parity tests; KV accumulation and the single head forward match CPU exactly). The tell was the failing draft steps returning **all-zero logits** (`metal_max=0.00`), sporadically winning the argmax at the last vocab row (248319/248320). Disabling the tape (`INFR_METAL_PROFILE=3`) made the divergence vanish entirely.

The bug: the decode-replay tape matches on a `(op-sequence, buffer-address)` fingerprint, which collides across the head's per-draft-step recompiles once the allocator reuses a freed buffer address → a structurally-stale tape replays garbage. Fixed in **#38** by invalidating the tape on `compile()` (the seam decode loop, which compiles once, keeps its replay — 0.6B tg128 unchanged at 145.6 t/s).

With #38 underneath, this branch now rebased on it:
- Metal MTP acceptance **0.26 → 0.816** (the CPU/f32 oracle is 0.96; the residual is ordinary f16-vs-f32, same as Vulkan).
- Still token-identical to greedy.

**Honest perf status:** MTP is still net-negative on Metal (12.3 vs 34.3 t/s decode) — but now for the same reason your own `INFR_MTP_TIME` verdict names on Vulkan: the verify is lm-head-dominated (74% here), and this model's vocab is 248320, so the per-draft-step lm_head GEMV is huge. That's the dispatch-fusion workstream you flagged, not a correctness issue. So: the port is correct, the alpha is right, and profitability waits on the lm_head cost — same ceiling as Vulkan.

Review order: **#38 first** (the general backend fix — stands alone, worth it independent of MTP), then this.

Issue #39, opened 2026-07-05
Metal MTP head-forward numerics — alpha 0.26 vs 0.95 (Vulkan f16) / 0.96 (CPU f32)


mxaddict:
Follow-up from #36 (merged), tracking the Metal-specific head-numerics gap so it doesn't live only in PR prose. Related tracking precedent: #30, #34.

**Symptom**: qwen35 MTP draft acceptance on Metal is α=0.26 vs Vulkan f16's ~0.95 and the CPU exact-f32 oracle's 0.96 (Qwen3.5-4B-MTP UD-Q4_K_XL, M3 Pro). Output stays CORRECT (the spec ≡ target-only-greedy invariant holds — verify catches the bad drafts), but at α=0.26 every reject pays the qwen35 no-rewind full reprefill, so `INFR_MTP=1` is net-negative on Metal until fixed.

**What's already ruled out** (from #36's isolation):
- Not the port logic: CPU driver (same core loop) hits 0.96.
- Not generic f16 rounding: Vulkan f16 hits ~0.95.
- Not the trunk: Metal spec-decode verify is token-identical (existing pinned test).
→ The divergence is inside the HEAD forward's Metal execution: the blk.{n_layer} attention layer at hd=256 (interleaved q+gate split, QkNormRope, sigmoid out-gate) and/or the eh_proj concat path, under Metal's kernels.

**Tools in place**:
- `generate_mtp_spec_cpu_timed` = the exact-f32 acceptance oracle to A/B against (lands in #36).
- Two new Metal parity tests from #36 (`qknormrope_hd256_partial_parity`, `attention_decode_hd256_short_kv_parity`) — both green on the M3, so the bug is in an op/shape combination they don't yet cover. Suggested next probes: the eh_proj CopyStrided-concat + Linear at [1, 2ne], the per-head interleaved q/gate strided views at hd256 on Metal, and a per-op h_mtp diff (CPU vs Metal, same inputs) through one head forward — the first op whose output diverges beyond f16 noise is the culprit.

**Bar**: Metal α within noise of Vulkan's (~0.95 on the same prompts), and `INFR_MTP=1` net-positive on the M3.

digitaloten:
Resolved — the gap was the replay tape (#38), not head numerics. Re-measured on M3 hardware against current `main`.

**Rig:** M3 Pro, macOS 26.5.2. Model: Qwen3.5-4B-MTP UD-Q4_K_XL (the same one from #36).

[...]

PR #42, opened 2026-07-06
feat(cli): measure Metal MTP self-speculation in bench

digitaloten:
The bench MTP arm (issue #33) only drove the Vulkan timed path, so `INFR_METAL=1 infr bench` on an MTP-head GGUF reported the baseline tg rate but no accept rate / phase split for Metal: `cmd_bench_metal` early-returns before the arm, and `bench_mtp_tg` hard-coded `generate_mtp_spec_vulkan_timed`.

**Change:** route `bench_mtp_tg` onto `generate_mtp_spec_metal_timed` under `INFR_METAL`, and add the same MTP arm to `cmd_bench_metal`. Non-MTP models and the Vulkan path are unchanged (`mtp = None` falls straight through to `print_bench_avg` — byte-identical output).

## Why — this closes out #39

Measuring the Metal accept rate is what #39 needed. On **Qwen3.5-4B-MTP UD-Q4_K_XL, M3 Pro**, current `main`:

```
tg128: 32.7 t/s | mtp128: 42.4 t/s (1.30x, alpha=0.95, draft 23% verify 64% catchup 12%)
tg256: 33.2 t/s | mtp256: 44.2 t/s (1.33x, alpha=0.97, draft 23% verify 64% catchup 13%)
tg128: 32.8 t/s | mtp128: 39.6 t/s (1.21x, alpha=0.95, draft 21% verify 68% catchup 11%)
```

Metal **alpha = 0.95–0.97**, matching Vulkan (0.95) and the CPU f32 oracle (0.96) — up from the **0.26** #39 tracked. The gap was the decode-replay tape fingerprint collision fixed in **#38**, not head numerics. MTP is now **net-positive on Metal (1.2–1.3×)** where it used to be net-negative, and stays **token-identical to greedy** (verified: `INFR_MTP=1` vs plain greedy at temp 0, byte-identical 200-token output).

I will post the same numbers to #39 and suggest closing it.

Verified locally on aarch64 (M3 Pro): `cargo fmt --all -- --check`, `cargo clippy -p infr-cli --all-targets --locked -- -D warnings`, and `RUSTFLAGS="-D warnings" cargo test -p infr-cli --no-run` all clean.

### DIFF
--- a/crates/infr-cli/src/main.rs
+++ b/crates/infr-cli/src/main.rs
@@ -1023,9 +1023,29 @@ fn bench_mtp_tg(
     let prompt = sentence.repeat(want.div_ceil(10));
     let mut samples = Vec::with_capacity(reps.max(1));
     let mut timing = infr_llama::mtp::MtpTiming::default();
+    // Backend selection mirrors the rest of the CLI: INFR_METAL routes the MTP driver onto the
+    // Apple-GPU trunk+head (issue #39 — measuring Metal's accept rate needs the Metal timed path,
+    // not Vulkan's), everything else stays on Vulkan. The two timed fns share a signature.
+    let use_metal = std::env::var("INFR_METAL").is_ok();
     for _ in 0..reps.max(1) {
-        let (stats, t) =
-            infr_llama::mtp::generate_mtp_spec_vulkan_timed(model, &head, &prompt, n_gen, |_| {})?;
+        let (stats, t) = if use_metal {
+            #[cfg(target_os = "macos")]
+            {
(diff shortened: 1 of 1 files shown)


mxaddict (approved):
Structure mirrors the Vulkan arm cleanly and closes #39's measurement gap — merging. One output nit your byte-identical claim misses: print_bench_avg_mtp's None branch hardcodes suffix "", so non-MTP models on Metal lose the " [metal]" tag this call previously passed. Pushing a suffix-param follow-up right after the merge.

Issue #35, opened 2026-07-05
qwen35 decode dispatch fusion — the lever for MTP parity and tg


mxaddict:
Spun out of #33's perf pass. The MTP spec cycle is now measured end-to-end (`INFR_MTP_TIME=1` + `INFR_PROF2=1`): the m=7 verify forward is **dispatch-count-bound** — 778 dispatches at ~11µs average (lm_head-family 51%, o/down projections 20%, deltanet 15%), costing 23ms vs 9.8ms for a single decode step. Graph build/compile is 0.0ms (pipeline cache), the no-rewind bookkeeping is correct (full-accept cycles run incremental), and reprefill is ~11% of verify time at α=0.95 — so the remaining gap to llama.cpp's own-speedup (theirs 1.68×, ours 1.18-1.21× on the same 7900 XTX) is per-dispatch overhead, not any single kernel.

The fix is **multi-layer/multi-op dispatch fusion** on the qwen35 decode/verify path — collapsing the many small per-layer linears (q/k/v/gate splits, alpha/beta projections, norms) into fewer, wider dispatches. This benefits ALL qwen35 decode (tg128 102 t/s vs llama 163 = 0.62× — same root cause), not just MTP.

Candidates, in rough order:
- Batch the per-layer `ssm_alpha`+`ssm_beta` F32 projections (2 dispatches × 24 layers) into one dispatch per layer or one for all layers (measured est. 1.5-2% alone — only worth it inside a broader fusion pass where the wins compound above bench noise ~3%)
- Fuse norm→projection chains (the AddBias/QkNormRope pattern precedent)
- One indirect multi-layer dispatch for the DeltaNet per-layer recurrence at decode (the batched-MoE routing prologue shows the indirect-dispatch pattern)
- lm_head at 51% of verify GPU time: the m=7 mrow GEMV over 248k vocab — check whether the batched GEMM tier beats mrow at m=7 for out_f this wide

Baselines to beat (4B MTP, 7900 XTX): `tg128 102 t/s | mtp128 120 t/s (1.18-1.21×, α=0.95, draft 31% / verify 57% / catchup 13%)`; llama.cpp same card: tg 163, mtp 273 (1.68×).

Related: the E2B "class-4 dispatch" note and the 2026-07-03 perf-state sweep (remaining gaps class-6/dispatch-bound) — same family of work.

digitaloten:
Candidate 4 (lm_head at m=7) answered on the Metal side — PR #43.

[...]

PR #43, opened 2026-07-06
perf(metal): route the wide lm_head off mrv onto cmm at m=2..8 (#35)

digitaloten:
One of #35's candidates: *"lm_head at 51% of verify GPU time — check whether the batched GEMM tier beats mrow at m=7 for out_f this wide."* Answer on Metal: **yes.**

The multi-row GEMV (`mrv`) path for m=2..8 K-quant Linears — speculative VERIFY's k+1 candidate rows, short suffix prefills — hoists weight bytes to registers and reuses them across the rows. It was tuned on the per-layer projections, whose `out_f` is ≤ 9216. The tied lm_head is `out_f = 248320` (vocab), 27× wider, where `mrv` spawns `out_f/2 * m` (~870k) simdgroups and loses to the `cmm` GEMM tile that streams the weight once at GEMM rate.

**Change:** gate the `mrv` path to `out_f < 65536` (comfortably above every per-layer projection, below any real vocab width), so the lm_head verify GEMV takes `cmm`. `INFR_METAL_LMHEAD_MRV` forces the old path (A/B + escape hatch).

## Measured (M3 Pro)

qwen3.5-4B-MTP UD-Q4_K_XL, MTP self-speculation (`INFR_METAL=1 infr bench -n 128`, needs the arm from #42), 3 clean alternating reps:

```
cmm (new): 46.1 / 46.0 / 45.9 t/s
mrv (old): 42.5 / 42.5 / 42.4 t/s     → +8.3% on the MTP cycle
```

Qwen3-0.6B self-speculation (vocab 151936, also ≥ 65536 → same reroute):

```
cmm (new): 127.7 / 127.8 tok/s
mrv (old): 124.4 / 124.1 tok/s        → +2.8%
```

So it generalizes to any large-`out_f` m=2..8 verify, not just qwen3.5.

## Correctness

Token-identical to plain greedy **and** to the old mrv path — verified `INFR_MTP=1` (default cmm) vs greedy and vs `INFR_METAL_LMHEAD_MRV=1` across four diverse prompts (prose, code, lists, haiku), byte-identical. Full `infr-metal` parity suite 94/94 on the M3. Alpha unchanged (0.95).

Verified locally on aarch64: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `RUSTFLAGS="-D warnings" cargo test -p infr-metal --no-run`, parity `--include-ignored` 94/94.

Note: the MTP bench numbers use the Metal MTP arm from #42 (open); the change itself is independent and also measurable via the spec-decode path above.

### DIFF
--- a/crates/infr-metal/src/exec.rs
+++ b/crates/infr-metal/src/exec.rs
@@ -1474,6 +1474,20 @@ impl MetalBackend {
                         "linear_q6k" => Some("linear_q6k_mrv"),
                         _ => None,
                     };
+                    // mrv's edge (weight bytes hoisted to registers, reused across the 2..8 rows)
+                    // was tuned on the per-layer projections (out_f <= 9216). The tied lm_head is
+                    // out_f = 248320 — 27x wider — where mrv spawns out_f/2 * m (~870k) simdgroups
+                    // and loses to the cmm GEMM tile, which streams the weight once at GEMM rate.
+                    // Measured on qwen3.5-4B-MTP verify (m=7): routing the lm_head to cmm is +8% on
+                    // the MTP cycle (1.30x -> 1.39x), token-identical to greedy. Gate at 65536 —
+                    // comfortably above every per-layer out_f, below any real lm_head/vocab width.
+                    // INFR_METAL_LMHEAD_MRV forces the old mrv path (A/B + escape hatch).
+                    let mr_kern =
+                        if out_f >= 65536 && std::env::var("INFR_METAL_LMHEAD_MRV").is_err() {
+                            None
+                        } else {
+                            mr_kern
+                        };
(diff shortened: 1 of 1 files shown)


mxaddict (approved):
Textbook slice: the missing out_f gate on Metal mrv mirrors exactly what the Vulkan mrv tuning found (m<=8 only pays below out_f<=8192 there), the token-identical verification against both greedy and the old path is the right bar for the MTP verify path, and the measured numbers + escape hatch follow house style. One non-blocking nit: env::var in the per-op dispatch — harmless at these op counts, but consider a OnceLock cache like the other INFR_ envs if you touch this again. Merging.

PR #44, opened 2026-07-06
feat(metal): native IQ2/IQ3 codebook kernels (all five formats)

digitaloten:
The IQ2/IQ3 codebook formats (IQ2_XXS, IQ2_XS, IQ2_S, IQ3_XXS, IQ3_S) had no native Metal kernel, so their weights dequanted to a cached f32 buffer — 4–8× the quantized size. On any checkpoint whose f32 blow-up exceeds the GPU working set, the OOM guard (#22) correctly refuses to run it — so these ultra-low-bit models (the ones people reach for to fit a big model in limited VRAM) could not run on Metal at all.

This adds native decode for **all five** IQ2/IQ3 formats.

## Change

One `DEC16_<fmt>` codebook unpacker per format (grid + sign lookup), wired through the existing `GEMV` / `RT` / `cmm` / `cmm_ks` / `hmm` kernel family via the shared macros — each format is one decode macro + five one-line instantiations. Plus the Linear execute + weight-load routing (native path instead of the f32 cache).

The grid + sign tables (`IQ2XXS/IQ2XS/IQ2S/IQ3XXS/IQ3S_GRID`, `KSIGNS_IQ2XS`) are **generated into the MSL source from `infr_core::iquant_grids`** (an `emit_grid` helper) — the same tables the CPU dequant reads — so the kernels are bit-exact by construction rather than by hand-transcribing 256–1024-entry tables, and there stays exactly one copy of each grid.

Format specifics: IQ2_XXS/IQ2_XS use the 7-bit KSIGNS table; IQ2_S and IQ3_S carry per-entry sign bytes; IQ2_S/IQ3_S pull extra grid-index bits from a `qh` plane; IQ3_S uses the `d*(1 + 2*nibble)` scale (the others `(0.5 + n)*0.25` / `*0.5`).

## Validation (M3 Pro)

- **15 parity tests** (GEMV / RT / cmm / hmm / w_off per format) bit-exact vs `dequant_block`. Full parity suite 94 → 103.
- The real `unsloth/Qwen3-4B-UD-IQ2_XXS` checkpoint (IQ2_XXS + IQ3_XXS + IQ2_S + IQ3_S tensors) now **runs on Metal at its full trained context with no f32 cache at all** — coherent and **token-identical to the CPU oracle**. Previously it OOM'd before the first forward; decode is 51 t/s (vs 37 on the partial-native path, which still f32-cached the residual formats).


mxaddict (approved):
Merging — this is the right shape end to end: grids emitted from infr_core::iquant_grids keep one source of truth and make the kernels bit-exact by construction, nearly half the diff is parity coverage (94→103), and the real-checkpoint token-identity vs the CPU oracle is the bar that matters. The capability unlock (IQ2/IQ3 checkpoints previously refused by the OOM guard now running natively at full context) is exactly what these formats exist for.

PR #70, opened 2026-07-14
fix(metal): thread RequestCtx through the macOS chat/seam path

mxaddict:
`main` does not compile on macOS.

The parallel-slots slice (`5f57f85`) added `req: Option<&RequestCtx>` to `ChatModel::generate`/`generate_constrained` and to the seam backend, and updated every caller **visible on Linux** — but `crates/infr-llama/src/chat/metal.rs` and the `generate_metal*` wrappers are `cfg(target_os = "macos")`, so clippy, `cargo test --workspace` and all 25 `gpu_seam` tests were green on Linux while the macOS build was broken.

The macOS CI job on the pending Metal PR stack is what surfaced it:
- `E0050`: method `generate` has 4 parameters but the trait declares 5 (`chat/metal.rs:95`, `:118`, `:202`)
- `chat/cpu.rs:49` already passed `req` into a 3-arg `generate_metal`

This threads `req` through the Metal chain properly rather than just satisfying the compiler, so **Metal serve honors per-request sampling like Vulkan does**: `generate_metal`, `generate_metal_session{,_constrained}`, `generate_dense_metal{,_session}`.

Greedy-only paths take `None` and say why: the MTP driver and the speculative decoder both require the target's argmax, so per-request sampling has nowhere to land — the same branches the Vulkan twin drops it on.

**This PR exists to get the macOS runner to verify it** — it cannot be checked on the Linux dev box.

### DIFF
--- a/crates/infr-llama/src/chat/metal.rs
+++ b/crates/infr-llama/src/chat/metal.rs
@@ -82,7 +82,7 @@ impl ChatModel for MetalSeamChat {
         // (No INFR_METAL_PROFILE suppression: the Metal backend reads it at CONSTRUCTION —
         // which happens inside this first generate — so unsetting it here would disable
         // profiling for the whole session, not just the warmup.)
-        self.generate("Hi", 2, &mut |_| {})?;
+        self.generate("Hi", 2, None, &mut |_| {})?;
         // Drop the warmup tokens so the first real prompt prefills clean slots from row 0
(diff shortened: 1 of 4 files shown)


mxaddict:
Superseded by #55, which landed the identical fix (same 4 files, same threading) while I was writing this. Closing as a duplicate — no need for two.

PR #90, opened 2026-08-03
fix(runtime): address correctness review findings

digitaloten:
## Summary

- reject malformed pipe-format arrays, unsolicited tool parsing, and zero attention-head metadata
- return loader errors for GGUF tensor-size overflow and snapshot GGUF bytes before exposing references
- publish shutdown state atomically and drop completed spin-pool results during panic unwinding
- add regression coverage and update runtime documentation

## Verification

- `cargo fmt --all -- --check`
- `cargo build --release --workspace --locked`
- `cargo test --workspace --locked`
- changed-package Clippy checks with `-D warnings`

The exact workspace Clippy gate reaches a pre-existing arm64 test warning at `crates/infr-cpu/src/kernels.rs:9757` (`TIER_TOL` is unused on this target). `cargo nextest` is not installed locally, so the full workspace suite was run with `cargo test` instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,24 @@
+# Changelog
+
+All notable changes to this project are documented in this file.
+
+The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
+
+## [Unreleased]
+
+### Changed
+
+- `Gguf::open` now snapshots files into immutable memory before exposing tensor
+  bytes, preventing later file changes from invalidating live references.
+
+### Fixed
+
+- Reject GGUF tensors whose encoded byte count overflows `usize` and model
+  metadata with zero attention heads.
+- Stop malformed pipe-format tool arrays from entering a non-progress allocation
+  loop.
+- Treat model JSON as a tool call only when the request offers a non-empty tool
+  list.
+- Publish graceful-shutdown state and its signal number atomically so
+  interrupted CLI commands retain the correct exit status.
+- Drop completed CPU spin-pool results when a sibling task panics.
--- a/crates/infr-chat/src/tools.rs
+++ b/crates/infr-chat/src/tools.rs
@@ -100,13 +100,13 @@ const MAX_VALUE_DEPTH: usize = 64;
 ///
 /// `s` has already had `<|"|>` replaced with `"`.
 ///
-/// Returns `None` — and only ever `None` — when nesting would exceed [`MAX_VALUE_DEPTH`];
-/// every other malformation stays lenient-by-design (partial objects, missing quotes and
-/// unterminated strings all yield a best-effort `Value`, as callers rely on). `None`
-/// propagates all the way out so the caller drops the whole call instead of acting on a
-/// truncated one. Note the degradation could NOT be "return `Value::Null` in place and stop
-/// consuming": the array arm loops `while` the cursor hasn't reached `]`, so a value that
-/// returns without advancing `i` spins forever pushing `Null`s — a hang and an OOM in place
+/// Returns `None` when nesting would exceed [`MAX_VALUE_DEPTH`] or when the current byte is
+/// a container delimiter that cannot begin a value. Other malformations stay lenient-by-design
+/// (partial objects, missing quotes and unterminated strings all yield a best-effort `Value`, as
+/// callers rely on). `None` propagates all the way out so the caller drops the whole call instead
+/// of acting on a truncated one. Note the degradation could NOT be "return `Value::Null` in place
+/// and stop consuming": the array arm loops `while` the cursor hasn't reached `]`, so a value
+/// that returns without advancing `i` spins forever pushing `Null`s — a hang and an OOM in place
 /// of the stack overflow. Propagating `None` is the only exit that terminates every loop.
 fn parse_value(s: &[u8], mut i: usize, depth: usize) -> Option<(Value, usize)> {
     // skip whitespace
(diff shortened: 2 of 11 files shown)


mxaddict:
Merged as `0aa0661`. Thanks — the tool-parser finding in particular is a good catch.

## Verified before merging

I reproduced the array hang rather than taking it on faith. On the pre-merge tree this model output allocates without bound:

```
<|tool_call>call:x{a:[}]}<tool_call|>
```

It reached 4.8 GB and aborted only because I ran it under `ulimit -v`. The bareword arm returns at `}` without advancing the cursor, so the array loop pushes empty strings forever. Since it is reachable from model output, any model that emits `[}` inside a tool call takes the server down — worth calling out as more than a parser tidy-up.

Local verification of the branch: 645 passed / 0 failed, `clippy --all-targets -D warnings` clean, `fmt --check` clean.

## One change reverted: the `Gguf::open` snapshot

Reverted in `5ba6b3f`. The hole it closes is real, but the cost is not affordable. Measured here on a 16 GiB Qwen3.6-27B, warm page cache, two reps each:

| | mmap | anonymous snapshot |
| --- | --- | --- |
| warm load | 1.87 s | **10.5 s** |
| anon RSS | 4.5 GiB | **20.2 GiB** |
| file RSS | 14.0 GiB | **0** |

Weights move from evictable page cache into anonymous memory, which is swap-only. Two consequences:

- The 5.6x load regression re-opens almost exactly the gap that earlier work closed.
- It is a capability loss, not just a slowdown. Page cache made "model larger than RAM" merely slow; anonymous memory makes it fatal. The Llama-4 Scout blob on this host is 47.5 GiB against 60 GiB of RAM. I did not run that one, for the obvious reason.

Nothing observes load time or RSS, so CI was green and would have stayed green.

Your underlying point stands, though, and the old `// SAFETY: the file is not modified while this Mmap is live` was exactly the unenforced-invariant shape you were reacting to. `Gguf::open` now says so explicitly instead of implying it is guaranteed, and the affordable direction — an advisory lock on the fd held for the mapping's lifetime — is written down with the two questions that block it: `flock` binds only cooperating processes, and an exclusive lock would stop two `infr` processes sharing one model, so a shared reader lock is probably the right shape. Tracked as B30 in `docs/backlog.md`, with the anonymous-copy measurements recorded so nobody re-proposes it blind.

Everything else from the PR was kept, including the tensor byte-count overflow check.

## Two nits, not blocking, filed as B31

- `SpinPool::collect` calls `out.as_mut_ptr()` twice — once for `CollectGuard`, once for `SendPtr`. The second reborrow pops the first pointer's tag under Stacked Borrows, and the guard then uses `slots` while unwinding. It will not miscompile, but it is UB by the model, and it is one line to fix: take the pointer once and derive both from it.
- `request_shutdown(0)` now returns `true` while latching nothing, since `compare_exchange(0, 0)` succeeds. Unreachable today, but it is public API in `infr-core`.

## One behaviour change worth a second look, filed as B32

Rejecting `,`/`}`/`]` in `parse_value` also changes valueless object entries, which the array fix did not require:

| input | before | after |
| --- | --- | --- |
| `{a:}` | `{"a": ""}` | call dropped |
| `{a:,b:1}` | `{"a": "", "b": 1}` | call dropped |

Well-formed input is unaffected, and dropping the call may well be the better behaviour — inventing an empty-string argument the model never wrote is its own bug. But the doc comment on `parse_value` still promises leniency for partial objects, so one of the two needs updating. Happy to take your view on which.

mxaddict:
Follow-up: all three nits from the review above are now fixed on `main` in `7784ab3`.

**`CollectGuard.slots`** — one `as_mut_ptr` now feeds both the tasks' writes and the unwind guard, so there is no second reborrow to invalidate the guard's pointer. Note the honest limit: this is fixed by construction and by reading, not verified by a tool. Miri is what would actually check it and this repo has never run it — the spin pool's real threads and spin loops, plus `infr-cpu`'s x86 intrinsics, make that a small project rather than a command. Filed as B31a so the gap is written down rather than implied away.

**`request_shutdown(0)`** — now a documented precondition with a `debug_assert`, placed before the compare-exchange so the release path the signal handler runs is still exactly one atomic instruction. Test asserts it; I confirmed the test fails if the assert is removed.

**`parse_value` doc vs behaviour** — I kept your stricter behaviour rather than restoring the old leniency. Binding `a` to an empty string when the model wrote no value is its own bug, and a dropped call is at least a visible failure. The doc now says that instead of promising uniform leniency, and a test pins where the line falls, since the distinction is easy to lose in a later edit:

- `{a:}` and `{a:,b:1}` — valueless, call dropped
- `{a:` — truncated, still parses best-effort to `{"a": null}`

That second case is the one that matters: a stream cut mid-value is the common malformation, and it still yields a call. If you'd rather valueless entries stayed lenient too, say so and I'll flip it — the test makes it a one-line change either way.

Workspace after all three: 647 passed / 0 failed, clippy and fmt clean.

mxaddict:
Correction to my last comment: the valueless-entry behaviour is now what it should have been from the start, in `37596df`.

Dropping the whole call was too blunt. One fumbled argument should not cost the arguments beside it — this parser exists to salvage imperfect model output, and discarding four good pairs because the fifth had no value is the wrong trade. The object arm now looks ahead past the colon and, on `,` or `}`, drops that ENTRY and carries on:

- `{a:,b:"text"}` → `{"b": "text"}` — the sibling survives
- `{a:1,b:,c:3}` → `{"a": 1, "c": 3}` — siblings on both sides survive
- `{a:}` → `{}` — call still fires, with no arguments
- `{a:` → `{"a": null}` — truncated, unchanged

Your rejection stays exactly where it is load-bearing: an array or the top level, where there is no entry to drop and no way to make progress. `{a:[}]}` still terminates.

Worth noting for anyone reading this later — the valueless-entry skip is a *second* place that has to advance the cursor correctly, i.e. a second chance at the original bug. It resumes ON the delimiter rather than past it, so the loop's own comma handling stays the single place the cursor moves.

To hold that down I added `delimiter_soup_always_terminates`: every 6-byte body over `{}[],:a"`, 262144 cases, asserting only that each call returns. Exhaustive rather than sampled, so it is a regression test rather than a fuzz run, and it costs 0.4 s. I checked it can actually fail — with your delimiter rejection removed it aborts on a 4.8 GB allocation, the same signature as the original bug.

Workspace: 648 passed / 0 failed, clippy and fmt clean.

PR #91, opened 2026-08-15
Add native Windows build and runtime support

Headmaster218:
Replace Unix-only file locking with Windows LockFileEx

Add Windows ranged file writes using seek_write

Add Windows-compatible blob linking fallback

Gate Unix fd APIs behind platform cfg

Keep Linux implementations unchanged

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1174,6 +1174,7 @@ dependencies = [
  "thiserror 2.0.18",
  "toml",
  "tracing",
+ "windows",
 ]
 
 [[package]]
--- a/crates/infr-core/Cargo.toml
+++ b/crates/infr-core/Cargo.toml
@@ -18,5 +18,8 @@ toml.workspace = true
 libc.workspace = true
 tracing.workspace = true
 
+[target.'cfg(windows)'.dependencies]
+windows = { version = "0.58", features = ["Win32_System_SystemInformation"] }
+
 [dev-dependencies]
 tempfile = "3"
--- a/crates/infr-core/src/hostmem.rs
+++ b/crates/infr-core/src/hostmem.rs
@@ -21,11 +21,8 @@
 /// a new allocation can have without swapping — it already accounts for reclaimable page cache, so
 /// it is exactly the figure this tier wants and not something derivable from `MemTotal`.
 ///
-/// **Every other platform answers `None` today**, and that is a gap rather than a design: macOS
-/// would need `host_statistics64`'s free/inactive/purgeable split and Windows
-/// `GlobalMemoryStatusEx`, neither of which is reachable through this workspace's existing
-/// dependencies, and neither of which could be verified on the machine this was written on. Adding
-/// either means adding a dependency, which is the user's call. Backlog B36.
+/// **Windows** reads `ullAvailPhys` from `GlobalMemoryStatusEx`. Every other platform answers
+/// `None` today; macOS would need `host_statistics64`'s free/inactive/purgeable split.
 ///
 /// **A cgroup memory limit overrides it.** `/proc/meminfo` is host-wide and knows nothing about the
 /// limit a container or a `systemd-run --scope -p MemoryMax=` puts on this process — measured on
@@ -41,10 +38,24 @@ pub fn available_bytes() -> Option<u64> {
             None => host,
         })
     }
-    #[cfg(not(target_os = "linux"))]
+    #[cfg(windows)]
+    {
+        Some(windows_memory_status()?.ullAvailPhys)
+    }
+    #[cfg(not(any(target_os = "linux", windows)))]
     {
         None
     }
+}
+
+#[cfg(windows)]
+fn windows_memory_status() -> Option<windows::Win32::System::SystemInformation::MEMORYSTATUSEX> {
+    use windows::Win32::System::SystemInformation::{GlobalMemoryStatusEx, MEMORYSTATUSEX};
+
+    let mut status = MEMORYSTATUSEX::default();
+    status.dwLength = std::mem::size_of::<MEMORYSTATUSEX>() as u32;
+    unsafe { GlobalMemoryStatusEx(&mut status).ok()? };
+    Some(status)
 }
 
 /// Memory this process may still commit before its cgroup kills it, or `None` when no ancestor
@@ -320,6 +331,22 @@ mod tests {
             * 1024;
         assert!(avail > 0, "available must be non-zero");
         assert!(avail <= total, "available {avail} exceeds total {total}");
+    }
+
+    #[cfg(windows)]
+    #[test]
+    fn the_windows_live_probe_is_plausible() {
+        let avail = available_bytes().expect("windows GlobalMemoryStatusEx should answer");
+        let status = windows_memory_status().expect("GlobalMemoryStatusEx");
+        assert!(
+            status.ullTotalPhys > 0,
+            "total physical memory must be non-zero"
+        );
+        assert!(
+            avail <= status.ullTotalPhys,
+            "available {avail} exceeds total {}",
+            status.ullTotalPhys
+        );
     }
 
     /// Headroom is the point: the budget never equals what is available, however much there is.
--- a/crates/infr-core/src/kernel_cache.rs
+++ b/crates/infr-core/src/kernel_cache.rs
@@ -427,6 +427,7 @@ mod tests {
     /// A pid that is guaranteed DEAD (spawned and reaped). Reusing a just-reaped pid is
     /// theoretically possible but the kernel hands out pids sequentially, so it will not happen
     /// within one test.
+    #[cfg(unix)]
     fn spawn_and_reap() -> u32 {
         let mut c = std::process::Command::new("true")
             .spawn()
@@ -517,6 +518,7 @@ mod tests {
 
     /// THE TRIPWIRE. A blob that hangs the GPU is perfectly well-formed, so no envelope check can
     /// see it — the only evidence is that the run which SEEDED from it never came back.
+    #[cfg(unix)]
     #[test]
     fn tripwire_discards_a_blob_whose_seeded_run_died() {
         let dir = tmp_dir("tripwire");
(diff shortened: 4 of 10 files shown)


Headmaster218:
Additional runtime validation on native Windows:

**Test environment**

* Windows 11
* AMD Radeon RX 7900 XTX 24 GB
* Vulkan backend
* MSVC Rust toolchain

**Build / test**

* `cargo build --release --locked -p infr-cli` ✅
* `cargo test --locked -p infr-hub --no-run` ✅
* `infr devices` correctly detects the RX 7900 XTX as the default Vulkan device ✅

**Real model inference**

* Ornith-1.0-9B Q6_K_XL:

  * `pp512`: ~2889 t/s
  * `tg256`: ~89.7 t/s
* Qwen3.6-35B-A3B UD-IQ4_NL_XL:

  * short-context `tg128`: ~151 t/s
  * 4K-context decode: ~143 t/s

So the Windows changes have been validated beyond compilation: model loading, Vulkan device selection, prefill, and token generation are all working successfully on native Windows.

Headmaster218:
Additional validation for the Windows host-memory detection change:

The Windows runtime can now detect available physical host memory automatically using `GlobalMemoryStatusEx`, so MoE paging no longer requires manually setting `INFR_DRAM_CACHE`.

**Implementation**

* Added a Windows implementation of `available_bytes()` using `GlobalMemoryStatusEx`
* Uses `ullAvailPhys` as the available physical-memory value
* Initializes `MEMORYSTATUSEX.dwLength` as required by the Win32 API
* Returns `None` on API failure, preserving the existing fallback behavior
* Existing Linux `/proc/meminfo` + cgroup headroom logic is unchanged

**Validation**

* `cargo build --release --locked -p infr-cli` ✅
* `cargo test --locked -p infr-core` ✅ — 183 tests passed
* Tested on Windows 11 + RX 7900 XTX

Without setting `INFR_DRAM_CACHE`, runtime logging now reports automatic host-tier sizing, for example:

```text
MoE host tier: sized automatically to 16.93 GB of ~45 GB available host memory
MoE host tier: 7 of 7 pool(s) paged, 16.93 GB arena of a 16.93 GB budget
```

This confirms that automatic DRAM host-tier sizing is working on native Windows while keeping `INFR_DRAM_CACHE` available as an explicit override.

Headmaster218:
latest results, runs with qwen35B A3B APEX-I-Balanced, 25GB model file


同配置基准 | 初始 16fbee2 | 里程碑 02e0bfb | 整体提升
-- | -- | -- | --
Prefill d0 / batch512 | 437.3 | 509.4 t/s | +16.5%
Prefill d100K / batch512 | 164.3 | 416.8 t/s | +153.7%，2.54x
Prefill d200K / batch512 | 89.1 | 299.8 t/s | +236.5%，3.36x
Decode d0 | 13.8 | 31.6 t/s | +129.0%，2.29x
Decode d100K | 20.5 | 47.7 t/s | +132.7%，2.33x
Decode d200K | 18.3 | 33.6 t/s | +83.6%，1.84x

Headmaster218:
## Qwen3.6-35B-A3B / RX 7900 XTX 当前性能

测试条件：

- Backend: Vulkan
- GPU: AMD Radeon RX 7900 XTX 24 GiB
- KV cache: Q8_0
- Prefill: 4096 tokens, 3 repetitions
- Decode: 1000 tokens
- UBatch: 4096
- Submit splitter: disabled/unlimited
- Context: synthetic depth
- Commit: `637f23f`

| Context depth | Expert cache | Prefill (t/s) | Decode (t/s) | Decode cache hit rate | 状态 |
|---:|---:|---:|---:|---:|---|
| 0K | 10 GiB | 2,901.4 | 93.7 | 约 97% | 已接近短上下文平台 |
| 32K | 12 GiB | 1,900.7 | 81.1 | 约 95% | 推荐配置 |
| 64K | 12 GiB | 1,358.2 | 67.2 | 约 93% | 推荐配置 |
| 128K | 12 GiB | 827.1 | 58.0 | 约 94% | Q8 最大安全配置 |
| 250K | 10 GiB | 476.3 | 43.7 | 约 94% | 长上下文推荐配置 |

### 累计优化成果

| 项目 | 初始基线 | 优化后里程碑 | 提升 |
|---|---:|---:|---:|
| Prefill, depth 0, batch 512 | 437.3 t/s | 509.4 t/s | +16.5% |
| Prefill, depth 100K, batch 512 | 164.3 t/s | 416.8 t/s | +153.7% / 2.54x |
| Prefill, depth 200K, batch 512 | 89.1 t/s | 299.8 t/s | +236.5% / 3.36x |
| Decode, depth 0 | 13.8 t/s | 31.6 t/s | +129.0% / 2.29x |
| Decode, depth 100K | 20.5 t/s | 47.7 t/s | +132.7% / 2.33x |
| Decode, depth 200K | 18.3 t/s | 33.6 t/s | +83.6% / 1.84x |
| Q8 Decode kernel chain, depth 200K | 31.55 t/s | 39.65 t/s | +25.7% |

### 当前结论

- 深上下文 Prefill 相对最初基线最高达到约 3.36x。
- 深上下文 Decode 相对最初基线达到约 1.84x。
- O(1) LRU 已消除单次 benchmark 中 14.9 亿至 21.1 亿次线性扫描。
- Q8 KV 相比 F16 节省 46.9% KV 显存。
- 250K 上下文推荐使用 Q8 KV + 10 GiB expert cache。
- 250K 下 Q8/10G 达到 476.3 t/s Prefill 和 43.7 t/s Decode。
- Prefill layer ring 已基本消除 6-14 GiB expert cache 容量对 Prefill 吞吐的影响。
- Decode 仍受 expert cache 命中率影响，长上下文增加 cache 的收益最明显。

mxaddict:
Thanks for the PR man, I'm reviewing it now 👍🏼

mxaddict:
Thanks for this — it's careful work, and the platform gating is done the way we'd want it (every new path behind `cfg(windows)`, Linux behaviour untouched). Notes below, with what I actually verified separated from what I'm only asking about.

## Verified: it does what it says

I cross-checked against `x86_64-pc-windows-gnu` locally, since nothing in CI builds Windows:

- `main` has **11 compile errors** for that target across `infr-core` + `infr-hub`. With this branch: **zero**, including `--all-targets` (so the test code compiles too).
- `cargo check --workspace --exclude infr-metal --target x86_64-pc-windows-gnu` is clean on this branch — `infr-vulkan`, `infr-llama`, `infr-cli` and `infr-server` all included, confirmed compiling rather than skipped.

A few details I checked because they're the ones that are usually wrong, and they're right here:

- The hand-written `Overlapped` matches the documented `OVERLAPPED` layout — two `ULONG_PTR`, the offset/pointer union, then `hEvent`.
- `LockFileEx` without `LOCKFILE_FAIL_IMMEDIATELY` blocks, which matches the existing `flock(LOCK_EX)` rather than quietly diverging from it. The `u32::MAX`/`u32::MAX` whole-range lock is the standard idiom.
- The `seek_write` loop is necessary and correct — Windows' positional write can be partial where `write_all_at` can't.
- `link_blob`'s hard-link fallback resolves the relative blob target against the link's parent, which is right.
- The `type RawFd = c_int` stubs in `p2p.rs`/`tp_sem.rs` look alarming next to a `libc::dup`, but `p2p_import` is gated on `p2p_supported(handle_type)` and no Windows driver exposes `VK_KHR_external_memory_fd`, so that path errors out loudly instead of running wrong. Worth a one-line comment saying exactly that, because the next reader will have the same scare I did.

## The failing check isn't yours

`cargo clippy` is red on `chunks_exact_to_as_chunks` in `crates/infr-gguf/src/dequant.rs`, which this branch doesn't touch. Rust 1.98 added that lint and CI runs clippy with `-D warnings` against current stable, so it went red on `main` too. Fixed in `f865e91` — a rebase onto `main` should clear it.

## The main gap: nothing verifies any of this

There's no Windows job in `ci.yml`, and `file_lock_is_exclusive` — the only test covering the lock — is now `#[cfg(not(target_os = "windows"))]`. So the new locking code has no coverage on the one platform it exists for, and "not supported here" would reach us as "passed".

The repo already has the pattern worth copying: the `metal-check` job cross-`cargo check`s an Apple target from an Ubuntu runner, purely to catch signature drift without paying for a macOS runner. A `windows-check` doing the same for `x86_64-pc-windows-msvc` would make this PR's central claim something CI defends rather than something a reviewer has to reproduce by hand. Adding it here would be very welcome.

## Questions rather than defects

- **Hand-rolled FFI vs the crate you're already adding.** The PR adds `windows = "0.58"` to `infr-core` for `GlobalMemoryStatusEx`, then declares `extern "system"` bindings plus a `#[repr(C)]` `OVERLAPPED` by hand in `infr-hub`. The same crate has these under `Win32_Storage_FileSystem`. Your transcription is correct, so this isn't a bug report — but a verified upstream definition can't drift and a hand-written ABI struct can. Was there a reason to avoid a second dependency edge on `infr-hub`?
- **`available_bytes` asymmetry.** The Linux arm deliberately picks `MemAvailable` over `MemFree`, then clamps by the cgroup limit, with a comment explaining that sizing an arena from the unclamped figure is an OOM kill in a container. The Windows arm returns `ullAvailPhys` with no equivalent clamp. Job Objects are the rough analogue — is the omission deliberate?
- The doc rewrite drops the `Backlog B36` pointer and the "could not be verified on the machine this was written on" caveat, replacing them with a flat claim. Given neither of us can run this, keeping a note about what's unverified would be worth more than the tidier sentence.

## Maintainer follow-up, not yours

There's no `CONTRIBUTING.md`, so there was no way for you to know either of these — I'll handle them:

- We keep a `CHANGELOG.md` under `## [Unreleased]`; this needs an entry.
- This PR makes several `docs/backlog.md` entries false rather than merely outdated: B30 says "`infr-hub` does not build there today", B58 says "`infr-hub` does not build on Windows at all", and B59 says `ranged::fetch_chunk` is "one more unix-only dependency in a crate that already does not build there". Our convention is that shipped entries get deleted, so those paragraphs should go with this change.

mxaddict:
I've merged your PR, thanks for the help on the windows work! 👍🏼

